PR #16, opened 2020-05-29
Mlsandar/feature/scenarios

logansandar:


src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -0,0 +1,74 @@
...
+from covid_model_seiir_pipeline.paths import (RegressionPaths, ODEPaths, ForecastPaths)
+from covid_model_seiir_pipeline.static_vars import COVARIATE_COL_DICT
+
+
+class ForecastDataInterface:
+
+    def __init__(self, forecast_root: Path, regression_root: Path, ode_fit_root: Path,
+                 location_file: Path):
+
+        self.forecast_paths = ForecastPaths(forecast_root)
```
mbannick, line 16: I like this interface a lot

src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -0,0 +1,74 @@
...
+    def load_covariate_scenarios(self, draw_id: int, location_id: int,
+                                 scenario_covariate_mapping: Dict[str, str]
+                                 ) -> pd.DataFrame:
+        # import file
+        scenario_file = self.regression_paths.get_scenarios_file(location_id=location_id,
+                                                                 draw_id=draw_id)
+        df = pd.read_csv(scenario_file)
+
+        # rename scenarios
+        missing_cols = set(list(scenario_covariate_mapping.keys())) - set(df.columns)
```
mbannick, line 48: @mlsandar Are you enforcing anywhere that you need to have exactly the same set of covariates in the forecast as it was in the regression? Not just that there aren't any missing ones.

src/covid_model_seiir_pipeline/paths.py
```diff
@@ -272,6 +273,10 @@ def get_location_dir(self, location_id: int):
             folder = matches[0]
         return self.root_dir / folder
 
+    def get_modelled_locations(self) -> List[int]:
```
mbannick, line 276: Would be better if we had Aaron write out a `metadata.yml` file in the future rather than scraping the folder names

src/covid_model_seiir_pipeline/forecasting/data.py
collijk: Will fix on integration testing

src/covid_model_seiir_pipeline/paths.py
collijk: I've been thinking about locations a bit.  Low priority, but good note.

PR #18, opened 2020-05-29
Feature/sequential option

njhenry:
Add the option to run sequential regression based on a `sequential` RegressionVersion setting that is `False` by default. Passes this option to an updated `ModelRunner.fit_beta_regression_prod()` method in the covid-model-seiir repository.

### DIFF
--- a/src/covid_model_seiir_pipeline/core/versioner.py
+++ b/src/covid_model_seiir_pipeline/core/versioner.py
@@ -11,7 +11,7 @@ BASE_DIR = Path('/ihme')
 
 # Dependency/input directories
 INPUT_DIR = BASE_DIR / 'covid-19/seir-inputs'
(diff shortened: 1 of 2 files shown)


src/covid_model_seiir_pipeline/core/versioner.py
```diff
@@ -336,17 +339,27 @@ class RegressionVersion(Version):
...
 
-    # Spline Arguments
+    # Spline required arguments
     degree: int
     knots: np.array
 
     # Regression Arguments
     covariates: Dict[str, Dict[str, Union[bool, List, float]]]
     covariates_order: List[List[str]] = None
     covariate_draw_dict: Dict[str, bool] = None
```
njhenry, line 349: The only new argument here is `sequential`. I also reordered some of the other RegressionVersion arguments into logical groups.

PR #26, opened 2020-06-08
add the avg_over parameter to keep the final scale of beta_shift.

zhengp0:


src/covid_model_seiir_pipeline/core/utils.py
```diff
@@ -127,19 +127,43 @@ def load_locations(directories):
...
+        beta_history = beta_fit[-avg_over:]
+        scale_final = beta_history.mean() / beta_pred[0]
+
     if window_size is not None:
         assert isinstance(window_size, int) and window_size > 0, f"window_size={window_size} has to be a positive " \
                                                                  f"integer."
-        scale = scale_init + (1 - scale_init)/window_size*np.arange(beta_pred.size)
-        scale[(window_size + 1):] = 1.0
+        scale = scale_init + (scale_final - scale_init)/window_size*np.arange(beta_pred.size)
+        scale[(window_size + 1):] = scale_final
```
collijk, line 166: Should we return the scale final as well?  I don't know what we're returning the scale init for.

zhengp0: That is for the diagnostic plots. For now I think it is good, since we are not plotting it yet :)

PR #28, opened 2020-06-10
Nathenry/feature/expose theta

njhenry:
Expose theta as a field in ForecastVersion. This also fixes a bug that caused errors when running forecasting separately from regression.


src/covid_model_seiir_pipeline/executor/beta_forecast.py
```diff
@@ -139,12 +139,19 @@ def run_beta_forecast(location_id: int, regression_version: str, forecast_versio
             gamma2=beta_params['gamma2'],
             N=N
         )
+        # If theta is set in the forecast settings, make it the same length as 
+        #  betas
+        if forecast_settings.theta is not None:
+            thetas = np.repeat(forecast_settings.theta, betas.size)
```
njhenry, line 145: `thetas` can be either None or a vector of the same length as `betas`. For now, we are keeping it as a constant in forecasting.

PR #31, opened 2020-06-10
Collijk/feature/uniform sampling of residual average

collijk:
Replace explicit parameterization with a uniform sample of the number of days to use to do the y-shift rescaling.


src/covid_model_seiir_pipeline/core/utils.py
```diff
@@ -59,7 +59,7 @@ def create_forecast_version(version_name, covariate_version,
     cache_version = cache_covariates(
         directories=directories,
         covariate_version=covariate_version,
-        location_ids=location_ids,
+        location_ids=location_ids,eta_shift_dict={'window_size': 14, 'avg_over': 30},
```
zhengp0: might be a typo `beta` or `eta`?

PR #34, opened 2020-06-12
Incorporating beta_shift updates

megan-h-moore:
According to ticket: https://jira.ihme.washington.edu/browse/GBDSCI-2689

I am not super familiar with working with the specs and dataclasses so I want to check that my assumption is correct that `window_size` and `avg_over` can be set to -1 as their default as a replacement for None (from the docs it seems like these have to be positive ints otherwise). If there is a better way to just set them to none let me know, because that is how they were being evaluated in the beta_shift function from this commit which is what I was basing this off of: https://github.com/ihmeuw/covid-model-seiir-pipeline/compare/10e137d..23843af#diff-e8f17e74967f7f40bfa9b6ec05b7d650R372

Going to test now

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/specification.py
+++ b/src/covid_model_seiir_pipeline/forecasting/specification.py
@@ -21,10 +21,12 @@ class ScenarioSpecification:
     """Forecasting specification for a scenario."""
     ALLOWED_ALGORITHMS = ('RK45',)
     NO_BETA_SCALING = -1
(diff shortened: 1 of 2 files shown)


megan-h-moore:
I believe in our discussions on Friday @collijk you said this might not be needed anymore or at least that the input arguments may have changed? Let me know and I will change the PR

collijk:
Handled separately in in forecast integration pr

PR #38, opened 2020-06-18
Collijk/feature/bye bye ode spline

collijk:


src/covid_model_seiir_pipeline/executor/beta_regression.py
```diff
@@ -60,7 +60,7 @@ def run_beta_regression(draw_id: int, regression_version: str):
         directories,
         covariate_version=settings.covariate_version,
         location_ids=location_ids,
-        draw_id = draw_id if any(settings.covariate_draw_dict.values()) else None
+        draw_id=draw_id if any(settings.covariate_draw_dict.values()) else None
```
miker985, line 63: no trailing comma?

collijk: Hah, my linter doesn't pick those up, so I didn't notice

PR #41, opened 2020-06-23
Collijk/integration/beta fit again

collijk:
Summary of major changes:
- Removed spline from beta fit.  This is now purely an ode fitting process.
- Port the relevant code from `covid-model-seiir` and drop the `ModelRunner` level of abstraction.

Integration test between current production pipeline code and this branch produce exact equality in the output dataframes.


specification_templates/fit.yaml
```diff
@@ -1,19 +1,10 @@
...
-  infection_version: '2020_06_12.10'
-  location_set_version_id: 708
+  infection_version: '2020_06_21.01'
+  location_set_version_id: 727
   output_root: ''
-  location_set_file: ''
+  location_set_file: '/ihme/covid-19/seir-pipeline-outputs/metadata-inputs/location_metadata_727.csv'
 parameters:
   n_draws: 10
-  degree: 3
```
collijk, line 8: We gutted the spline piece of the ode fit, so these options are no longer necessary.

src/covid_model_seiir_pipeline/ode_fit/model.py
```diff
@@ -0,0 +1,389 @@
+# -*- coding: utf-8 -*-
```
collijk, line 1: Straight port of https://github.com/ihmeuw/covid-model-seiir/blob/develop/src/covid_model_seiir/ode_model/ode_process.py

src/covid_model_seiir_pipeline/ode_fit/specification.py
```diff
@@ -25,16 +22,6 @@ class FitParameters:
     """Specifies the parameters of the ODE fit."""
     n_draws: int = field(default=1000)
 
-    degree: int = field(default=3)
```
collijk, line 28: No more spline

src/covid_model_seiir_pipeline/ode_fit/task.py
```diff
@@ -51,34 +47,23 @@ def run_ode_fit(draw_id: int, ode_version: str):
         gamma1=fit_specification.parameters.gamma1,
         gamma2=fit_specification.parameters.gamma2,
         solver_dt=fit_specification.parameters.solver_dt,
-        spline_options={
```
collijk, line 54: No more spline

src/covid_model_seiir_pipeline/ode_fit/task.py
```diff
@@ -51,34 +47,23 @@ def run_ode_fit(draw_id: int, ode_version: str):
...
-        spline_space=fit_specification.parameters.spline_space,
+
     )
 
-    # ----------------------- BETA SPLINE + ODE -------------------------------- #
-    # Start a Model Runner with the processed inputs and fit the beta spline / ODE
-    mr = ModelRunner()
-    mr.fit_beta_ode(beta_fit_inputs)
+    # Convert infections into beta and fit compartments to data.
+    ode_model = model.ODEProcess(beta_fit_inputs)
```
collijk, line 55: From the port of https://github.com/ihmeuw/covid-model-seiir/blob/develop/src/covid_model_seiir/ode_model/ode_process.py.  Don't use model runner anymore.  Use the `ODEProcess` interface directly.

src/covid_model_seiir_pipeline/ode_fit/model.py
```diff
@@ -0,0 +1,389 @@
...
+    def create_start_end_date_df(self):
+        df_result = pd.DataFrame({
+            self.col_loc_id: self.loc_id,
+            'start_date': self.start_date,
+            'end_date': self.end_date
+        }, index=[0])
+        return df_result
+
+
+@dataclass
```
miker985, line 271: feels so good not having to write this `__init__`

src/covid_model_seiir_pipeline/ode_fit/model.py
```diff
@@ -0,0 +1,389 @@
...
+        Args:
+            input: ODEProcessInput
+        """
+        self.df_dict = input.df_dict
+        self.col_date = input.col_date
+        self.col_cases = input.col_cases
+        self.col_pop = input.col_pop
+        self.col_loc_id = input.col_loc_id
+        self.col_lag_days = input.col_lag_days
+        self.col_observed = input.col_observed
```
miker985, line 304: `df_dict` and `col_date` (and it appears the rest - but I only CTRL + F'd the first two) are only used within this function. Do they need to be member variables instead of just assigning `input` to `self`?

collijk: They do not.  This whole module has a bunch of stuff that we'll want to clean up and simplify.  Right now the ODEProcessInput is a pretty useless struct.  It's packed, passed in, then immediately unpacked.

PR #48, opened 2020-06-26
Collijk/integration/beta fit location handling

collijk:
The primary work here is to get location handling working properly.  

How this works:
- Default behavior: if no location info is specified, we use all locations modeled by the infectionator.
- In the fit specification there is a `location_set_version_id` key and a `location_file` key. 
  - If both are specified, we fail
  - if  `location_set_version_id` is specified, we pull from GBD, subset to most detailed, intersect with locs available from infectionator (warning about missing locations)
  - if `location_file` is specified, we read in the file, subset to most detailed, intersect with locs available from infectoinator (warning about missing locations)
- At the command line there is a `-l` or `--location-specification` argument
  - Defaults to None
  - If an int is specified, the fit specification is updated to use as the `location_set_version_id`
  - If a path is specified, the fit specification is updated to use as the `location_file`

The main function loads the location list from primary sources and then writes them out to a `locations.yaml` in the run directory.  The individual tasks read the location list from the run directory.

Other things here:
- Some bugfixes in the cli arg handling
- Simplification of some of the object usage (swapping custom objects for standard types where reasonable) to better facilitate testing.

**Validation**: Ran and passed integration notebook (exact equality with last integration test outputs produced by the production code).

### DIFF
--- a/src/covid_model_seiir_pipeline/cli.py
+++ b/src/covid_model_seiir_pipeline/cli.py
@@ -68,7 +68,7 @@ def fit(run_metadata,
     cli_tools.configure_logging_to_files(run_directory)
     main = cli_tools.monitor_application(ode_fit.do_beta_fit,
                                          logger, with_debugger)
-    app_metadata, _ = main(fit_spec, run_directory)
+    app_metadata, _ = main(fit_spec)
 
     run_metadata['app_metadata'] = app_metadata.to_dict()
     run_metadata.dump(run_directory / 'metadata.yaml')
--- a/src/covid_model_seiir_pipeline/cli.py
+++ b/src/covid_model_seiir_pipeline/cli.py
@@ -29,7 +29,7 @@ def fit(run_metadata,
         fit_specification,
         infection_version,
         location_specification,
-        output_root, mark_dir_as_best, production_tag,
+        output_root, mark_best, production_tag,
         verbose, with_debugger):
     """Runs a beta fit on a set of infection data."""
     cli_tools.configure_logging_to_terminal(verbose)
@@ -72,7 +72,7 @@ def fit(run_metadata,
     run_metadata['app_metadata'] = app_metadata.to_dict()
     run_metadata.dump(run_directory / 'metadata.yaml')
 
-    cli_tools.make_links(app_metadata, run_directory, mark_dir_as_best, production_tag)
+    cli_tools.make_links(app_metadata, run_directory, mark_best, production_tag)
 
     logger.info('**Done**')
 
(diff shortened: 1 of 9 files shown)


src/covid_model_seiir_pipeline/cli.py
```diff
@@ -24,7 +24,7 @@ def seiir():
               help="Either a location set version id used to pull a list of"
                    "locations to run, or a full path to a file describing"
                    "the location set.")
-@cli_tools.add_output_options(default_output_root=DEFAULT_SEIIR_OUTPUT_ROOT)
+@cli_tools.add_output_options(paths.SEIR_FIT_OUTPUTS)
```
miker985, line 27: `DEFAULT_SEIIR_OUTPUT_ROOT` can be safely deleted now, right?

src/covid_model_seiir_pipeline/ode_fit/data.py
```diff
@@ -1,35 +1,63 @@
...
+        directory.
 
         """
-        desired_locs = pd.read_csv(self.location_metadata_file)["location_id"].tolist()
-        modeled_locs = self.infection_paths.get_modelled_locations()
-        missing_locs = list(set(desired_locs).difference(modeled_locs))
-        if missing_locs:
+        modeled_locations = self.infection_paths.get_modelled_locations()
+        if desired_locations is None:
+            desired_locations = modeled_locations
```
miker985, line 45: `return modeled_locations` would more efficient and *arguably* clearer depending on how you feel about multiple `return` statements

src/covid_model_seiir_pipeline/ode_fit/workflow.py
```diff
@@ -66,9 +58,9 @@ def __init__(self, fit_specification: FitSpecification):
             resume=True
         )
 
-    def attach_ode_tasks(self):
-        for draw_id in range(self.fit_specification.parameters.n_draws):
-            task = self._ode_fit_task_template.get_task(draw_id, self.fit_specification.data.output_root)
+    def attach_ode_tasks(self, n_draws: int):
```
miker985, line 61: on the one hand, i strongly prefer explicit

on the other hand, now it seems there's redundant data being passed in since it's still available from the fit spec

how do you decide when you to use the more error-prone implicit internal data vs using explicit parameters that are, arguably, redundant?

src/covid_model_seiir_pipeline/cli.py
collijk: Oh, yes

src/covid_model_seiir_pipeline/ode_fit/data.py
collijk: I'm usually not a fan. I find code with single entry and exit points easier to reason about.  The performance cost we're paying here is pretty trivial. :shrug:

src/covid_model_seiir_pipeline/ode_fit/workflow.py
collijk: The workflow no long hangs on to the fit specification.  It only gets the info it needs (a version) on init.

collijk: So no redundancy.  I'm making the specifications abstractions for use by cli/main/task code only as a way to communicate parameters.  Executor code and modeling code will get primative types whereever possible (or potentially a data container where necessary).

PR #49, opened 2020-06-26
First attempt at using travis

miker985:
WIP. Will update when Travis starts passing (that's the whole point of this PR)


collijk:
Shell script install works, don't need this one anymore

PR #50, opened 2020-06-26
Install via pure bash script

miker985:
WIP

### DIFF
--- a/ci/travis_before_install
+++ b/ci/travis_before_install
@@ -12,11 +12,12 @@ if [[ ! -f miniconda/bin/activate ]]; then
 fi
 source "$PWD/miniconda/bin/activate"
 
+command -v pip
+command -v python
+
 # install gfortran - which is not installed by limetr but required for it to work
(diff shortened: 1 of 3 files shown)


miker985:
This is taking significantly longer than before so I'm hoping it actually works. Adding you (James) because of that

ci/travis_before_install
```diff
@@ -0,0 +1,22 @@
...
+
+command -v conda || { echo conda not found ; exit 1; }
+if [[ ! -f miniconda/bin/activate ]]; then
+    echo No activate file;
+    exit 1;
+fi
+source "$PWD/miniconda/bin/activate"
+
+# install gfortran - which is not installed by limetr but required for it to work
+conda install -c conda-forge fortran-compiler
```
miker985, line 16: per our conversation - the problem was not that conda wasn't activating correctly with the above call. the problem is that travis doesn't have gfortran installed by default, but the IHME cluster does.

miker985:
Thanks for running with this

PR #52, opened 2020-06-29
Test ode fit

miker985:
This is a proof-of-concept that tests `run_ode_fit`

It tests almost entirely with mock objects. Fundamentally what it does is ensure a bunch of calls to `ODEDataInterface` are made with correct arguments (normally from `ode_model`)

### DIFF
--- a/src/covid_model_seiir_pipeline/ode_fit/model.py
+++ b/src/covid_model_seiir_pipeline/ode_fit/model.py
@@ -83,8 +83,7 @@ class SingleGroupODEProcess:
         df.sort_values(self.col_date, inplace=True)
         date = pd.to_datetime(df[col_date])
         try:
-            end_date = self.today + np.timedelta64(self.day_shift -
-                                                   self.lag_days, 'D')
+            end_date = self.today + np.timedelta64(int(self.day_shift - self.lag_days), 'D')
         except ValueError:
             msg = f"""end_date debug:
                 self.today:           {self.today!r} ({type(self.today)})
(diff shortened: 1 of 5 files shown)


tests/ode_fit/test_model.py
```diff
@@ -0,0 +1,109 @@
...
+            'beta': 0.310171949194388,
+            'S': 3682620.648469025,
+            'E': 1668.104871971228,
+            'I1': 921.078728218612,
+            'I2': 572.4297222704859,
+            'R': 56163.725831125324,
+            'newE': 445.045338509743,
+            'newE_obs': 445.045338509743,
+        }
+        assert r2 == e2
```
collijk, line 72: Test the last day in the data as well.  Operationally, it's the same, but it's fit on data from a short term forecast rather than observations.

src/covid_model_seiir_pipeline/ode_fit/model.py
```diff
@@ -82,8 +82,9 @@ def __init__(self, df,
         self.lag_days = lag_days
         df.sort_values(self.col_date, inplace=True)
         date = pd.to_datetime(df[col_date])
-        end_date = self.today + np.timedelta64(self.day_shift -
-                                               self.lag_days, 'D')
+        # cast this from numpy.int64 to int because travis/tox was erroring on
+        # this, only for python 3.7, with a ValueError
+        end_date = self.today + np.timedelta64(int(self.day_shift - self.lag_days), 'D')
```
miker985, line 87: this is soo dirty

PR #53, opened 2020-07-01
Collijk/integration/remove forecast diagnostics

collijk:
Doing forecast integration in stages so y'all don't have a single very complex PR to review.  I'm treating the forecast stage as a non-working scaffold.  I'll incrementally fill in the main function as things work.

Here:
- Adjusting how the cli calls the main to be consistent with fit/regress.  spec will be dumped in the main
- Removing diagnostics tasks.  They've been replaced by plotting code later in the pipeline


src/covid_model_seiir_pipeline/forecasting/main.py
```diff
@@ -2,50 +2,35 @@
...
-    forecast_wf = ForecastWorkflow(forecast_specification, location_ids)
-    forecast_wf.attach_scenario_tasks()
-    forecast_wf.attach_splice_tasks()
-    forecast_wf.attach_scaling_diagnostics_tasks()
-    forecast_wf.attach_forecast_diagnostics_tasks()
-    forecast_wf.run()
+    # forecast_wf = ForecastWorkflow(forecast_specification, location_ids)
+    # forecast_wf.attach_scenario_tasks()
+    # forecast_wf.attach_splice_tasks()
+    # forecast_wf.run()
```
miker985, line 36: doesn't this make `seiir forecast` almost a no-op as written?

PR #62, opened 2020-07-09
Add scenario paths

collijk:
Keep the top level path interface with multiple scenario paths underneath generated from the specification.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/data.py
+++ b/src/covid_model_seiir_pipeline/forecasting/data.py
@@ -21,7 +21,9 @@ class ForecastDataInterface:
 
     @classmethod
     def from_specification(cls, specification: ForecastSpecification) -> 'ForecastDataInterface':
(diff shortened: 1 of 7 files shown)


src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -21,7 +21,9 @@ def __init__(self,
 
     @classmethod
     def from_specification(cls, specification: ForecastSpecification) -> 'ForecastDataInterface':
-        forecast_paths = paths.ForecastPaths(specification.data.output_root, read_only=False)
+        forecast_paths = paths.ForecastPaths(specification.data.output_root,
+                                             read_only=False,
+                                             scenarios=list(specification.scenarios))
```
miker985, line 26: What type are `scenarios` if not a list?

collijk: Dict[str, ScenarioSpecification]

PR #64, opened 2020-07-09
Extract beta hat computation, clean up regressors a bit

collijk:
Okay.  So.  There is one bit of modeling code shared between the forecast and regression stages and that's the computation of beta hat from coefficients and covariates.  This is literally just element-wise matrix multiplication and then a sum over columns.

This PR does one important thing:

- Pulls that computation out of the regressors/predict function and into a shared function.

This PR does two less important things, because I was already there and trying to make sense of the code:
- Makes a tiny adapter around the CovModelSet to bridge our specification abstraction and the abstraction in slime.
- Refactor the regressor classes in three ways:
  - Strip out a lot of unneccesary debug code
  - Give those classes a literal common interface.
  - Favor functional application over imperative `model.fit` calls that attach new data pieces to the regressor objects.

There's still some work to be done on these (e.g. the `BetaRegressorSequential.fit` method consumes one of its attributes, which seems problematic for testing), but it's deep in the stack and not super important right this very second.

### DIFF
--- a/src/covid_model_seiir_pipeline/model_runner.py
+++ b/src/covid_model_seiir_pipeline/model_runner.py
@@ -1,3 +1,5 @@
+import pandas as pd
+
 from covid_model_seiir_pipeline.forecasting.model import ODERunner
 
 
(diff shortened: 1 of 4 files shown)


src/covid_model_seiir_pipeline/model_runner.py
```diff
@@ -41,3 +43,33 @@ def forecast(model_specs, init_cond, times, betas, thetas=None, dt=0.1):
         """
         forecaster = ODERunner(model_specs, init_cond, dt=dt)
         return forecaster.get_solution(times, beta=betas, theta=thetas)
+
+
+# FIXME: The only "modeling" code shared between the stages.  Where to put it?
```
miker985, line 48: `from covid_model_seiir_pipeline.math import compute_beta_hat`?

collijk: yeah, something like that.

PR #65, opened 2020-07-10
Collijk/integration/forecasting task

collijk:
Okay, several pieces here.  At a high level this produces forecasted SEIIR components from the outputs of the regression stage and some covariate predictions.  Important things to note:

- Splice has been carved off.  It'll now be part of it's own "post-processing" stage
- You can specify a different covariate pool than was used in regression. It'll warn you that something might be fishy, but otherwise assume you know what you're doing.  It still has some hard constraints about the covariates that must be present. This radically simplifies the tracking behavior for covariates.
- Forecast has been reoriented to run by draw.  This makes it consistent with the regression stage and vastly reduces file count.

Validation was done by looking at the relative changed in the data points from a run on the production pipeline across all locations/dates/draws.  X axis is in percent change and y axis is count of data points.

**Scaling parameters**

image

**SEIR compartments**

image

Interpretation:  All errors here are symmetric about 0 (unbiased) and small and correspond very closely with what's expected based on the regression integration testing.


src/covid_model_seiir_pipeline/forecasting/model.py
```diff
@@ -111,37 +110,81 @@ def get_solution(self, times, beta, theta=None, solver="RK4"):
...
+    scale_init = beta_fit_final / beta_hat_start
+
+    beta_past = beta_past.reset_index().set_index(['location_id', 'date'])
+    log_beta_resid = np.log(beta_past['beta'] / beta_past['beta_pred']).rename('beta_resid')
+    scale_final = np.exp(log_beta_resid
+                         .groupby(level='location_id')
+                         .apply(lambda x: x.iloc[-avg_over:].mean()))
+
+    import pdb
+    pdb.set_trace()
```
miker985, line 169: WIP?

PR #72, opened 2020-08-11
Add base classes for task templates and workflows to reduce boilerplate

collijk:
This PR pulls out base classes for the task templates and workflows.  

Context: I'm about to start adding new tasks and maybe a new workflow and wanted to reduce some boiler plate.

### DIFF
--- a/src/covid_model_seiir_pipeline/workflow_template.py
+++ b/src/covid_model_seiir_pipeline/workflow_template.py
@@ -1,0 +1,54 @@
+import abc
+from typing import Dict
+
+from jobmon.client import Workflow, BashTask
+from jobmon.client.swarm.executors.base import ExecutorParameters
+from jobmon.client.swarm.workflow.task_dag import DagExecutionStatus
+
+from covid_model_seiir_pipeline import utilities
+
+
+class TaskTemplate:
+
+    task_name_template: str = None
+    command_template: str = None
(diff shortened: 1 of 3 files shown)


src/covid_model_seiir_pipeline/workflow_template.py
```diff
@@ -0,0 +1,54 @@
...
+    @classmethod
+    def get_task(cls, *_, **kwargs) -> BashTask:
+        task = BashTask(
+            command=cls.command_template.format(**kwargs),
+            name=cls.task_name_template.format(**kwargs),
+            executor_parameters=cls.params,
+            max_attempts=1
+        )
+        return task
+
```
megan-h-moore, line 26: Jobmon Guppy has the concept of a Task Template (as well as Tools and a few other new objects) that get implemented a little bit differently than the above so I was thrown off for a second thinking you were using those. But now that I see you are creating your own using the original BashTask, so I am glad we are on the same wavelength and the need is there!

If you are interested in using Guppy I am happy to send docs or walk you through the changes. See [https://hub.ihme.washington.edu/display/DataScience/Jobmon+Release+Notes](url) for starters

collijk: Definitely interested!

I'll steal some of your time next week and we can do the upgrade!

PR #73, opened 2020-08-11
Collijk/feature/residual average mean shift

collijk:
This is the first half of a refactor of the beta residual averaging.  

Context: We've pivoted the whole workflow to run by draw.  Last week (or before) CJLM, asked for a methods update to adjust the way the final scaling works in locations with a small number of deaths.  This requires knowledge of the mean of the residuals across all draws.  In the production pipeline (under duress), I accomplished this by doing an extra run of the SEIR pipeline, and then manually building a dataset of total deaths and residual means by location.  This PR offloads that task to a preprocessing step before the forecasting stage runs.  

The next PR will make adjustments to the residual scaling algorithm to remove all the work done here and work from the dataset produced by this stage instead.


src/covid_model_seiir_pipeline/forecasting/workflow.py
```diff
@@ -5,6 +5,27 @@
 
 from covid_model_seiir_pipeline.workflow_template import TaskTemplate, WorkflowTemplate
 
+FORECAST_RUNTIME = 1000
+FORECAST_MEMORY = '5G'
+FORECAST_CORES = 1
+FORECAST_SCALING_CORES = 50
+FORECAST_QUEUE = 'd.q'
```
miker985, line 12: Consider pulling these from environment variables, using the above values as defaults

I've never been happy in the long-run when I have to update code to use a less busy queue

PR #74, opened 2020-08-12
Collijk/feature/use scaling parameters in forecast

collijk:
Second have of the beta residual scaling.  See #73 for the first half.  

This PR removes the pieces of the beta scaling algorithm from the forecast task and uses the data computed in the preprocessing step.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/data.py
+++ b/src/covid_model_seiir_pipeline/forecasting/data.py
@@ -96,11 +96,11 @@ class ForecastDataInterface:
     def load_total_deaths(self):
         """Load cumulative deaths by location."""
         # TODO: add metadata to regression and forecast paths.
-        with (self.forecast_paths.root_dir / 'metadata.yaml').open() as metadata_file:
+        with (self.regression_paths.root_dir / 'metadata.yaml').open() as metadata_file:
             metadata = yaml.full_load(metadata_file)
 
         # TODO: metadata abstraction?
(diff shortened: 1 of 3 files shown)


src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -96,14 +96,15 @@ def check_covariates(self, scenarios: Dict[str, ScenarioSpecification]):
...
+        with (self.regression_paths.root_dir / 'metadata.yaml').open() as metadata_file:
             metadata = yaml.full_load(metadata_file)
 
         # TODO: metadata abstraction?
-        model_inputs_version = metadata['regression_metadata']['infectionator_metadata']['death']['metadata']['model_inputs_metadata']['output_path']
+        model_inputs_version = metadata['infectionator_metadata']['death']['metadata']['model_inputs_metadata']['output_path']
         full_data_path = Path(model_inputs_version) / 'full_data.csv'
         full_data = pd.read_csv(full_data_path)
         total_deaths = full_data.groupby('location_id')['Deaths'].max().rename('deaths').reset_index()
+        total_deaths['location_id'] = total_deaths['location_id'].astype(int)
```
miker985, line 107: What were the location_id's loading as?

collijk: floats :-(

collijk: It's not a big deal until you put them into the index and want to look them up using `.loc` or `.at`, and then pandas get's very finicky.

PR #75, opened 2020-08-13
Collijk/feature/postprocessing

collijk:
This brings a post-processing step to replace the collapse/combine pipeline stage and have it run automatically after the forecast has finished.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/task/postprocessing.py
+++ b/src/covid_model_seiir_pipeline/forecasting/task/postprocessing.py
@@ -24,6 +24,18 @@ def run_seir_postprocessing(forecast_version: str) -> None:
 
     for scenario in forecast_spec.scenarios:
         infections, deaths, r_effective = load_output_data(scenario, data_interface)
+        cum_deaths = deaths.groupby(level='location_id').cumsum()
+        max_deaths = cum_deaths.groupby(level='location_id').max()
+        upper_deaths = max_deaths.quantile(.975, axis=1)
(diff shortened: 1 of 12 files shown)


src/covid_model_seiir_pipeline/forecasting/task/postprocessing.py
```diff
@@ -0,0 +1,155 @@
...
+import numpy as np
+import pandas as pd
+
+from covid_model_seiir_pipeline import static_vars
+from covid_model_seiir_pipeline.forecasting.specification import ForecastSpecification
+from covid_model_seiir_pipeline.forecasting.data import ForecastDataInterface
+
+log = logging.getLogger(__name__)
+
+LOAD_OUTPUT_CPUS = 3
```
miker985, line 18: Is this a place where you want to scale based off of `fthread`s the job has?

`ihmeutils.cluster.max_cpus()` can do that for you.

miker985: Oh, I just realized this needs to build outside IHME. Here's the code if it matters - https://stash.ihme.washington.edu/users/miker985/repos/ihmeutils/browse/src/ihmeutils/cluster.py

collijk: Cool!

I need to come back and properly profile at the end (see todo about reconsidering the usage of multiprocessing).  I wasn't thinking of making this dynamic, but that's a neat package.

PR #76, opened 2020-08-14
Collijk/feature/beta postprocessing

collijk:
This adds the beta time series for the past and future and a time series of the residuals of log beta in the past to the outputs.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
+++ b/src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
@@ -106,7 +106,7 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str):
                     .sort_values(['location_id', 'date'])
                     .set_index(['location_id']))
     components['theta'] = thetas.reindex(components.index).fillna(0)
-    data_interface.save_components(components, scenario_name, draw_id)
+    data_interface.save_components(components.reset_index(), scenario_name, draw_id)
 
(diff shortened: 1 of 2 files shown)


src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -106,7 +106,7 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str):
                     .sort_values(['location_id', 'date'])
                     .set_index(['location_id']))
     components['theta'] = thetas.reindex(components.index).fillna(0)
-    data_interface.save_components(components, scenario_name, draw_id)
+    data_interface.save_components(components.reset_index(), scenario_name, draw_id)
```
miker985, line 109: I kind of wish `components = components.reset_index()  # restore location_id column prior to save` was it's own line

collijk: The thing is, I don't want to reset the index for the computations below, just for saving.

miker985: Maybe add a comment then? I find it very unintuitive that you're serializing something that is **not quite** what you're using to process data.

collijk: Fair.  Will add some comments in the next PR.  I need to rework this section to allow for multiple algorithms anyway.

PR #78, opened 2020-08-18
Collijk/feature/separate resampling from postprocessing

collijk:
Sorry, this PR got away from me.  This does:

- Introduces a step after all the forecasts run to build a resampling map for draws.  The context: based on the cumulative deaths in the worse scenario, the top and bottom 2.5% (configurable) are dropped from the data and replaced with random samples of the other draws.  These same draws need to be resampled from all other measures to produce a coherent set of outputs (which we definitely weren't doing before.  
- Parallelizes postprocessing by scenario.  This is still the longest step of the forecasting process and needs to be parallelized by measure as well (but like, later).
- Adds covariates and their coefficients to the outputs.


src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -169,6 +170,14 @@ def load_covariate_info(self, covariate: str, info_type: str, location_ids: List
             index_columns.append('date')
         return info_df.set_index(index_columns)
 
+    def save_raw_covariates(self, covariates: pd.DataFrame, scenario: str, draw_id: int):
+        self.forecast_marshall.dump(covariates, key=MKeys.forecast_raw_covariates(scenario=scenario, draw_id=draw_id))
+
+    def load_raw_covariates(self, scenario: str, draw_id: int):
+        covariates = self.forecast_marshall.load(key=MKeys.forecast_raw_covariates(scenario=scenario, draw_id=draw_id))
```
mbannick, line 177: I like the `forecast_marshall` :)

src/covid_model_seiir_pipeline/forecasting/postprocessing_lib.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+    with multiprocessing.Pool(FORECAST_SCALING_CORES) as pool:
+        resampled = pool.map(_runner, measure_data)
+    return resampled
+
+
+def resample_draws_by_measure(measure_data: pd.DataFrame, resampling_map):
+    output = []
+    for location_id, loc_map in resampling_map.items():
+        loc_data = measure_data.loc[location_id]
```
mbannick, line 172: Do you need a `measure_data.loc[location_id].copy()` here?

src/covid_model_seiir_pipeline/forecasting/postprocessing_lib.py
```diff
@@ -0,0 +1,181 @@
...
+        loc_data[loc_map['to_resample']] = loc_data[loc_map['to_fill']]
+        loc_data = pd.concat({location_id: loc_data}, names=['location_id'])
+        if isinstance(loc_data, pd.Series):
```
miker985, line 175: What causes this check to need to exist?

collijk: Pandas being infuriating and not giving me the thing I want.

collijk: If you call `df.loc[single_value_indexer]` on a dataframe with a multi-index, it looks in the outermost level of the index, plucks out the sub_df corresponding to the indexer, and drops the outermost level from the index.  Hence me adding it back in on 174.  If you do the same thing on a dataframe with a normal index and the indexer corresponds to one row, it coerces the row to a series.  

In this situation, most of our data is (location_id, date), but we have some dataframes that are just by location.

PR #82, opened 2020-08-31
Collijk/feature/multiple mandate reimposition

collijk:
This expands the mandate reimposition loop to run until all locations are below the death threshold until the end of the forecast period.  This takes a thing that used to cost me 10-12 hours and makes it happen in about 10 minutes.


src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -165,6 +152,7 @@ def parse_arguments(argstr: Optional[str] = None) -> Namespace:
 
 
 def main():
+    configure_logging_to_terminal(1)
```
miker985, line 155: Can this `1` be set as a kwarg or otherwise made less cryptic?

PR #84, opened 2020-09-03
Collijk/feature/national level aggregates

collijk:
Perform aggregation of all measure to the global and national level for downstream consumers.


ctroeger:
James, any chance you can have it aggregate to World Bank regions, too? Location_set_id 111, location_set_version_id 775

src/covid_model_seiir_pipeline/forecasting/postprocessing_lib.py
```diff
@@ -311,3 +319,96 @@ def resample_draws(measure_data: pd.DataFrame, resampling_map: Dict[int, Dict[st
     resampled = pd.concat(output)
     resampled.columns = [f'draw_{draw}' for draw in resampled.columns]
     return resampled
+
+
+def sum_aggregator(measure_data: pd.DataFrame, hierarchy: pd.DataFrame, _population: pd.DataFrame) -> pd.DataFrame:
```
miker985, line 324: `_population` does not appear to be used here

PR #85, opened 2020-09-04
Collijk/feature/new pipeline proportional theta

collijk:
This changes the interpretation of theta so that it's relative to the size of the susceptible/exposed population size.  Corresponding updates have been made to the r effective calculation.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
+++ b/src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
@@ -59,6 +59,11 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str):
     # We'll need this to compute deaths and to splice with the forecasts.
     infection_data = data_interface.load_infection_data(draw_id)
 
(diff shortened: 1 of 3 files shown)


src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -59,6 +59,11 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str):
     # We'll need this to compute deaths and to splice with the forecasts.
     infection_data = data_interface.load_infection_data(draw_id)
 
+    if ((1 < thetas) | thetas < -1).any():
+        raise ValueError('Theta must be between -1 and 1.')
+    if (beta_params['sigma'] - thetas).any():
```
miker985, line 64: The error message implies this should be `((beta_params['sigma']) - thetas >= 1).any()`. What am I missing?

collijk: Nothing.  good catch.

miker985 (approved):

PR #89, opened 2020-09-13
Collijk/refactor/split beta forecast model

collijk:
Working on mean level mandate reimposition and just doing some organizing first.  This PR is almost all code motion.  I'll leave a comment where it's not.

this does:
- Removes integration notebooks.  The seiir grid plots have been updated and so those are being used for integration testing now.
- Changes the 'mandate_reimposition' algorithm choice to 'draw_level_mandate_reimposition' as we're about to have two flavors.
- Turns `model.py` into a subpackages split across three modules.  I'm almost out of the scramble to make it work phase and I'm trying to figure out the fault lines for good data structure definitions.  The split here is into the actual ode forecast, the mandate reimposition calculation, and the output metric computation.


src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -70,11 +70,13 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str):
     future_components = model.run_normal_ode_model_by_location(initial_condition, beta_params, betas, thetas,
                                                                location_ids, scenario_spec.solver)
     logger.info('Processing ODE results and computing deaths and infections.')
-    components = model.splice_components(past_components, future_components)
-    components['theta'] = thetas.reindex(components.index).fillna(0)
-    infections, deaths, r_effective = model.compute_output_metrics(infection_data, components, beta_params)
+    components, infections, deaths, r_effective = model.compute_output_metrics(infection_data,
```
collijk, line 73: This is the only real code change. `compute_output_metrics` now does the components splicing.

PR #97, opened 2020-09-18
Collijk/feature/workflow specification from config

collijk:
Sketching out the design to connect the workflow specification up to the configuration system.  This is not tested yet.

The goal here is to allow an additional configuration block `workflow` in the forecast and regression specifications that'll allow the user to, e.g., change which queue we run on or to fiddle with runtime or memory for pipeline tasks.

### DIFF
--- a/src/covid_model_seiir_pipeline/regression/specification.py
+++ b/src/covid_model_seiir_pipeline/regression/specification.py
@@ -12,16 +12,17 @@ from covid_model_seiir_pipeline.workflow_template import (
 
 class RegressionTaskParams(ExecutorParameters):
 
-    def __init__(self,
-                 queue,
-                 max_runtime_seconds=None,
-                 m_mem_free='2G',
-                 num_cores=1, **kwargs):
-        super().__init__(num_cores=num_cores,
-                         queue=queue,
-                         max_runtime_seconds=max_runtime_seconds,
-                         m_mem_free=m_mem_free,
-                         **kwargs)
+    @property
+    def default_cores(self) -> int:
+        return 1
+
+    @property
+    def default_memory(self) -> str:
+        return '2G'
+
+    @property
+    def default_runtime(self) -> int:
+        return 3000
 
 
 class RegressionWorkflowSpecification(WorkflowSpecification):
--- a/src/covid_model_seiir_pipeline/regression/specification.py
+++ b/src/covid_model_seiir_pipeline/regression/specification.py
@@ -2,6 +2,36 @@ from dataclasses import dataclass, field
 from typing import Dict, Tuple, List
 
 from covid_model_seiir_pipeline.utilities import Specification, asdict
+from covid_model_seiir_pipeline.workflow_template import (
+    ExecutorParameters,
+    WorkflowSpecification,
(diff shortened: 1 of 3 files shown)


src/covid_model_seiir_pipeline/regression/workflow.py
```diff
@@ -1,29 +1,29 @@
...
+    @property
+    def workflow_name_template(self) -> str:
+        return 'seiir-regression-{version}'
+
+    def build_task_templates(self, task_specifications: Dict[str, ExecutorParameters]):
+        return {
+            'regression': TaskTemplate(
+                task_name_template="beta_regression_draw_{draw_id}",
+                command_template=(
+                        f"{shutil.which('beta_regression')} " +
```
mbannick, line 20: WOW I wish I had known this `shutil` function existed...

src/covid_model_seiir_pipeline/workflow_template.py
```diff
@@ -2,50 +2,128 @@
...
+    def to_dict(self):
+        return {
+            'project': self.project,
+            'task_specifications': {
+                task: spec.to_dict() for task, spec in self.task_specifications.items(),
+            }
+        }
+
 
 class TaskTemplate:
```
mbannick, line 68: Are you going to transition to using the guppy version of jobmon?

src/covid_model_seiir_pipeline/regression/specification.py
```diff
@@ -2,6 +2,37 @@
...
+    def default_cores(self) -> int:
+        return 1
+
+    @property
+    def default_memory(self) -> str:
+        return '2G'
+
+    @property
+    def default_runtime(self) -> int:
+        return 3000
```
miker985, line 25: `return 3000  # seconds`

src/covid_model_seiir_pipeline/workflow_template.py
```diff
@@ -2,50 +2,128 @@
...
+        super().__init__(num_cores=num_cores,
+                         queue=queue,
+                         max_runtime_seconds=max_runtime_seconds,
+                         m_mem_free=m_mem_free,
+                         **kwargs)
+
+    @abc.abstractmethod
+    @property
+    def default_runtime(self) -> int:
+        ...
```
miker985, line 37: Fun fact: this line would be valid PERL, and it would do the equivalent of `raise NotImplementedError()` if it was ever run by the interpreter

src/covid_model_seiir_pipeline/workflow_template.py
```diff
@@ -2,50 +2,128 @@
...
+
         stdout, stderr = utilities.make_log_dirs(version, prefix='jobmon')
 
         self.workflow = Workflow(
             workflow_args=self.workflow_name_template.format(version=version),
-            project="proj_covid",
+            project=workflow_spec.project,
             stderr=stderr,
             stdout=stdout,
             seconds_until_timeout=60*60*24,
```
miker985, line 103: Do you foresee any use in setting this to something shorter? My understanding of your covid workflow is a day is **way** too long for things to sit and you'd need to e.g., find another queue if it was taking this long

PR #111, opened 2020-11-18
Collijk/feature/vaccine updates

collijk:
Sorry this is a big messy PR.  It's unifying three attempts at this from two people as requirements were changing.  

Nearly all the work of interest is in beta_forecast.py and the modules it calls functions from: ode_forecast.py and forecast_metrics.py.


specification_templates/forecast.yaml
```diff
@@ -4,9 +4,9 @@ data:
   output_root: ''
 scenarios:
   worse:
-    algorithm: 'normal' # normal, mean_level_mandate_reimposition, draw_level_mandate_reimposition
```
collijk, line 7: Removed all code related to mean level mandate reimposition.

specification_templates/forecast.yaml
```diff
@@ -4,9 +4,9 @@ data:
...
+    algorithm: 'normal' # normal, draw_level_mandate_reimposition
     solver: 'RK45'
-    system: 'new_theta'
```
collijk, line 9: Removed 'old_theta' now that 'new_theta' has been in prod for a couple of months.

specification_templates/vaccine.yaml
```diff
@@ -1,19 +1,19 @@
 data:
-  regression_version: '2020_09_15.01'
-  covariate_version: '2020_09_10.03'
+  regression_version: '2020_11_15.02'
+  covariate_version: '2020_11_15.02'
```
kathrynline, line 3: I'm obviously not an expert on this codebase. Are all of these yaml files hand-entered?

collijk: Yes.  In lieu of having about 30 cli args, the options are all specified here.

src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -23,21 +23,38 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str, *
...
+    # We'll use the same params in the ODE forecast as we did in the fit.
+    beta_params = data_interface.load_beta_params(draw_id=draw_id)
     # Thetas are a parameter generated from assumption or OOS predictive
     # validity testing to curtail some of the bad behavior of the model.
     thetas = data_interface.load_thetas(scenario_spec.theta)
+    if ((1 < thetas) | thetas < -1).any():
+        raise ValueError('Theta must be between -1 and 1.')
+    if (beta_params['sigma'] - thetas >= 1).any():
+        raise ValueError('Sigma - theta must be smaller than 1')
+
```
miker985, line 35: I'm sure I've seen this check before. Any idea if that PR supersedes this or if it's duplicated logic?

collijk: This just moves the check

miker985 (approved):
Gentle request to method parameter lists in how you define and call them.

Basically just 1 arg per line and always include trailing commas for the last arg.

You do this is many places but not always, and it creates a lot of noise when things are being updated.

PR #112, opened 2020-11-25
Collijk/feature/vaccine scenarios

collijk:


src/covid_model_seiir_pipeline/forecasting/model/forecast_metrics.py
```diff
@@ -89,17 +90,17 @@ def _cleanup(infections: pd.Series) -> pd.DataFrame:
                 .sort_index()['infections'])
 
     susceptible_columns = [c for c in components.columns if 'S' in c]
-    exposed_and_protected_columns = [c for c in components.columns if 'E_p' in c]
+    newE_protected_columns = [c for c in components.columns if '_p' in c and 'S' not in c]
```
miker985, line 93: `newE` is ... new exposures?

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
```diff
@@ -197,6 +199,7 @@ def __init__(self,
         self.gamma1 = constants.gamma1
         self.gamma2 = constants.gamma2
         self.N = constants.N
+        self.system_params = constants.system_params
```
miker985, line 202: Suggest using a `.copy()` of this since it's a mutable dict

src/covid_model_seiir_pipeline/forecasting/model/forecast_metrics.py
collijk: New exposures among those in the protected group.

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
collijk: Good idea

src/covid_model_seiir_pipeline/forecasting/model/forecast_metrics.py
collijk: I'll add some comments.  This is a confusing thing.

miker985 (approved):

PR #115, opened 2020-11-30
Collijk/feature/infection elastispliner regression

collijk:
This is the first PR (of probably 3) to incorporate the merge of the infectionator and elastispliner that Ryan has put together.  This one handles the regression pipeline stage.

Stuff done here:
- Update the CLI pathing and metadata to reflect the new data source
- Data handling for new format (1 file per draw vs 1 file per loc-draw)
- Change `cases` to `infections` wherever appropriate to reflect the correct terminology for the data.
- Fix usage of the builtin name `input`
- Update logging tools in the regression task.

### DIFF
--- a/src/covid_model_seiir_pipeline/regression/data.py
+++ b/src/covid_model_seiir_pipeline/regression/data.py
@@ -76,7 +76,7 @@ class RegressionDataInterface:
 
         """
         # Fixme: Hack.
-        draw_0_data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0))
+        draw_0_data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0), usecols=['location_id'])
         modeled_locations = draw_0_data['location_id'].unique().tolist()
 
         if desired_locations is None:
@@ -105,7 +105,7 @@ class RegressionDataInterface:
         infection_inputs = pd.read_csv(self.infection_paths.get_model_data_file())
         pop = infection_inputs.groupby('location_id')['population'].max().reset_index()
 
-        data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0))
+        data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=draw_id))
         data = data[data['location_id'].isin(location_ids)]
         data['date'] = pd.to_datetime(data['date'])
         data = data.merge(pop, on='location_id', how='left')
--- a/src/covid_model_seiir_pipeline/regression/data.py
+++ b/src/covid_model_seiir_pipeline/regression/data.py
@@ -75,7 +75,10 @@ class RegressionDataInterface:
         directory.
 
         """
-        modeled_locations = self.infection_paths.get_modelled_locations()
+        # Fixme: Hack.
+        draw_0_data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0))
+        modeled_locations = draw_0_data['location_id'].unique().tolist()
+
         if desired_locations is None:
             desired_locations = modeled_locations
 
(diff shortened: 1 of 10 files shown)


src/covid_model_seiir_pipeline/regression/data.py
```diff
@@ -75,7 +75,10 @@ def filter_location_ids(self, desired_locations: List[int] = None) -> List[int]:
         directory.
 
         """
-        modeled_locations = self.infection_paths.get_modelled_locations()
+        # Fixme: Hack.
+        draw_0_data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0))
```
miker985, line 79: `read_csv` takes a `usecols` arg that will let you specify just `location_id`

src/covid_model_seiir_pipeline/regression/data.py
```diff
@@ -98,21 +101,16 @@ def dump_location_ids(self, location_ids: List[int]):
...
-                logger.warning(f'Negatives found in infectionator inputs for location id {loc}.  Dropping.')
-                continue
-            dfs[loc] = loc_df
-
-        return dfs
+    def load_infection_data(self, draw_id: int, location_ids: List[int]) -> pd.DataFrame:
+        infection_inputs = pd.read_csv(self.infection_paths.get_model_data_file())
+        pop = infection_inputs.groupby('location_id')['population'].max().reset_index()
+
+        data = pd.read_csv(self.infection_paths.get_infection_file(draw_id=0))
```
miker985, line 108: Shouldn't this be `draw_id=draw_id`?

src/covid_model_seiir_pipeline/regression/model/ode_fit.py
```diff
@@ -328,33 +328,23 @@ def __init__(self, input):
...
+                self.col_infections,
+                self.col_pop,
+                self.col_loc_id,
+                day_shift=(self.day_shift,)*2,
+                lag_days=self.lag_days,
+                alpha=(self.alpha,)*2,
+                sigma=(self.sigma,)*2,
+                gamma1=(self.gamma1,)*2,
+                gamma2=(self.gamma2,)*2,
+                solver_class=RK4,
```
miker985, line 344: There's a `solver` param in the YAML files. Is it not pertinent to this?

src/covid_model_seiir_pipeline/regression/task.py
```diff
@@ -1,68 +1,74 @@
...
-        col_pop=INFECTION_COL_DICT['COL_POP'],
-        col_loc_id=INFECTION_COL_DICT['COL_LOC_ID'],
-        col_lag_days=INFECTION_COL_DICT['COL_ID_LAG'],
-        col_observed=INFECTION_COL_DICT['COL_OBS_DEATHS'],
+        col_date='date',
+        col_infections='infections_draw',
+        col_pop='population',
+        col_loc_id='location_id',
+        col_lag_days='duration',
+        col_observed='observed_deaths',
```
miker985, line 41: What led to moving away from `INFECTION_COL_DICT`?

src/covid_model_seiir_pipeline/regression/task.py
```diff
@@ -1,68 +1,74 @@
...
     data_df = pd.concat(location_data.values())
-    data_df['date'] = pd.to_datetime(data_df['date'])
     regression_betas = beta_hat.merge(covariates, on=['location_id', 'date'])
     regression_betas = beta_fit.merge(regression_betas, on=['location_id', 'date'], how='left')
-    merged = data_df.merge(regression_betas, on=['location_id', 'date'], how='outer').sort_values(['location_id', 'date'])
-    merged = merged[(merged['obs_deaths'] == 1) | (merged['deaths_draw'] > 0)]
+    merged = (data_df
+              .merge(regression_betas, on=['location_id', 'date'], how='outer')
+              .sort_values(['location_id', 'date']))
+    merged = merged[(merged['observed_deaths'] == 1) | (merged['deaths_draw'] > 0)]
```
miker985, line 71: I _think_ @JosephWagner would write this as `merged = merged.loc[..., :]`. I'm still not clear why to use one over the other.

Or maybe I'm just wrong about Joe

src/covid_model_seiir_pipeline/regression/task.py
```diff
@@ -97,7 +105,7 @@ def parse_arguments(argstr: Optional[str] = None) -> Namespace:
 
 
 def main():
-
+    configure_logging_to_terminal(verbose=1)  # Debug level
```
miker985, line 108: Recommend adding a ["count" verbosity arg](https://click.palletsprojects.com/en/7.x/options/#counting) to CLI arguments

src/covid_model_seiir_pipeline/regression/data.py
collijk: Good catch

src/covid_model_seiir_pipeline/regression/model/ode_fit.py
collijk: There's not one in the regression spec.  There are several inconsistencies present around the regression parameterization and it's relationship to the forecast that are on my to do list.

src/covid_model_seiir_pipeline/regression/task.py
collijk: It's a half-baked abstraction.  There's a lot of inconsistent use of indirection in the regression code.  This is a struct that unpacks two other structs that itself exists simply to be unpacked again immediately in the line 51 call. 

I'm not planning on sticking with raw strings in the long run.  Just want to strip out some abstraction so I can plan more easily.

src/covid_model_seiir_pipeline/regression/task.py
collijk: That's a thing I would do.  I tend to use .loc and .iloc for everything.  Not mucking with logic here though.

src/covid_model_seiir_pipeline/regression/task.py
collijk: I agree in the long run. I have a plan for structured logging of the SEIR tasks.  Right now, just trying to make consistent with the forecast tasks.

miker985 (approved):

PR #122, opened 2020-12-04
Collijk/feature/screen and vaccinate

collijk:
This PR adds a configuration parameter picked up from the specification that says some proportion of vaccines are guaranteed to be delivered to susceptible individuals.  This corresponds to a test and treat strategy vs. our existing blind vaccination strategy.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
+++ b/src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
@@ -259,11 +259,11 @@ class _VaccineSEIIR(ODESystem):
         n = sum(unvaccinated)
 
         v_total = v_non_efficacious + v_efficacious
-        efficacy = v_efficacious / v_total
         screened_vaccines = p_screening * v_total
         unscreened_vaccines = v_total - screened_vaccines
         s_vaccines = screened_vaccines + s/n * unscreened_vaccines
         if s_vaccines:
+            efficacy = v_efficacious / v_total
             rho_unprotected = 1 - efficacy
             rho_protected = efficacy * (1 - p_immune)
             rho_immune = efficacy * p_immune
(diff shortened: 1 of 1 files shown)


src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
```diff
@@ -249,7 +249,7 @@ def __init__(self, *args, **kwargs):
 
     def _group_system(self, t: float, y: np.array, infectious: float,
```
kathrynline, line 250: Is there a docstring for these variables? Looking at this codebase for the first time, I'm confused.

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
```diff
@@ -259,47 +259,36 @@ def _group_system(self, t: float, y: np.array, infectious: float,
...
-        # Effective vaccines are efficacious vaccines delivered to
-        # susceptible, unvaccinated individuals.
-        v_effective = s/n * v_efficacious
-        # Some effective vaccines confer immunity, others just protect
-        # from death after infection.
-        v_immune = p_immune * v_effective
-        v_protected = v_effective - v_immune
-
-        # vaccinated and unprotected come from all bins.
-        # Get count coming from S.
```
kathrynline, line 271: What is S?

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
kathrynline: What is "screened" in general in this script?

kathrynline (approved):

collijk:
this turned out to be an interesting experiment, but not something work keeping around.

PR #123, opened 2020-12-04
Collijk/feature/workflow specification from config take 2

collijk:
The main point of this PR is to allow specification of the queue, project, and task execution parameters for the SEIR pipeline in the model specification files.  To do this, I've introduced a pair of specification abstractions that do the same work as the other specifications: providing sensible defaults and validating specification parameters.  One for the tasks and one for the overall workflow.

I've also done some refinement of the existing abstractions around the actual workflow and workflow task implementations to reduce some boilerplate and to make use of the new specification abstractions.

### DIFF
--- a/src/covid_model_seiir_pipeline/forecasting/specification.py
+++ b/src/covid_model_seiir_pipeline/forecasting/specification.py
@@ -181,6 +181,7 @@ class ForecastSpecification(Specification):
     def parse_spec_dict(cls, forecast_spec_dict: Dict) -> Tuple:
         """Construct forecast specification args from a dict."""
         data = ForecastData(**forecast_spec_dict.get('data', {}))
+        workflow = ForecastWorkflowSpecification(**forecast_spec_dict.get('workflow', {}))
         postprocessing = PostprocessingSpecification(**forecast_spec_dict.get('postprocessing', {}))
         scenario_dicts = forecast_spec_dict.get('scenarios', {})
         scenarios = []
@@ -188,7 +189,7 @@ class ForecastSpecification(Specification):
             scenarios.append(ScenarioSpecification(name, **scenario_spec))
         if not scenarios:
             scenarios.append(ScenarioSpecification())
-        return tuple(itertools.chain([data, postprocessing], scenarios))
+        return tuple(itertools.chain([data, workflow, postprocessing], scenarios))
 
     @property
     def data(self) -> ForecastData:
--- a/src/covid_model_seiir_pipeline/forecasting/specification.py
+++ b/src/covid_model_seiir_pipeline/forecasting/specification.py
@@ -1,17 +1,61 @@
 from dataclasses import dataclass, field
 import itertools
-from typing import Dict, Tuple, Union
+from typing import Dict, NamedTuple, Tuple, Union
 
 from covid_model_seiir_pipeline.utilities import Specification, asdict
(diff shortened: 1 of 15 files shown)


collijk:
src/covid_model_seiir_pipeline/workflow_tools/template.py should compare to the old src/covid_model_seiir_pipeline/workflow_template.py.  Sorry about the code motion.  I didn't realize I'd introduced jobmon dependencies into the test suite until I made the PR.

src/covid_model_seiir_pipeline/forecasting/task/postprocessing.py
```diff
@@ -9,14 +9,18 @@
 import yaml
 
 from covid_model_seiir_pipeline import static_vars
-from covid_model_seiir_pipeline.forecasting.specification import ForecastSpecification, ScenarioSpecification
+from covid_model_seiir_pipeline.forecasting.specification import (
+    ForecastSpecification,
+    ScenarioSpecification,
+    FORECAST_JOBS
```
miker985, line 15: trailing comma

src/covid_model_seiir_pipeline/forecasting/specification.py
```diff
@@ -1,17 +1,67 @@
...
+class ForecastWorkflowSpecification(WorkflowSpecification):
+    """Specification of execution parameters for forecasting workflows."""
+
+    tasks = {
+        FORECAST_JOBS.scaling: ScalingTaskSpecification,
+        FORECAST_JOBS.forecast: ForecastTaskSpecification,
+        FORECAST_JOBS.sentinel: SentinelTaskSpecification,
+        FORECAST_JOBS.resample: ResampleTaskSpecification,
+        FORECAST_JOBS.postprocess: PostprocessingTaskSpecification,
+    }
```
miker985, line 64: what is this mapping for? It looks like expected types but it's a class constant

collijk: It get's used here: https://github.com/ihmeuw/covid-model-seiir-pipeline/pull/123/files#diff-b3cc748ab1833dfcf96f29d680ca4968ab2352f8ab0fac8986eb649d074ad191R123-R139
to turn the TaskSpecification subclasses (which are essentially bundles of sensible default values) into TaskSpecification objects which resolve inputs from the specification file with the defaults and do a little validation.

collijk: I'm doing this all with class variables to mirror the way other specifications are constructed out of data classes with a very similar interface.

src/covid_model_seiir_pipeline/workflow_tools/specification.py
```diff
@@ -0,0 +1,148 @@
...
+    Subclasses are meant to inherit and provide default values for the
+    class variables. At runtime those defaults will be used to fill values
+    not provided in the specification file when the specification file
+    is parsed and then the full task specification will be validated against
+    the values in the `validate` method of this class.
+
+    """
+    default_max_runtime_seconds: int
+    default_m_mem_free: str
+    default_num_cores: int
```
miker985, line 31: Why is this an ABC?

Recommend `__init_subclass__` for ensuring these defaults are filled in by subclasses.

collijk: It's an abc because it has no defaults and therefore shouldn't be used directly. Your solution is better though.  I always forget about that hook.

PR #124, opened 2020-12-07
Collijk/feature/optimize ode

collijk:
Adding in the vaccination pieces caused the ``beta_forecast`` stage to increase in duration by a factor of 10.  The ode forecast, in particular, went from something taking about 3 minutes to something taking about 25 minutes.  The structure of the ode makes it particularly amenable to cython style optimization (The process is fundamentally a series of nested loops in a way that can't be vectorized or parallelized).  I thought I'd try out the new kid on the optimization block, numba, first.  Turned out to be super easy and give me more than enough speedup that going the cython route even as a comparison was not worth the time.

This PR has the absolute minimum required to incorporate the optimizations.  I made an adaptor `_ODERunnerOptimized` that parallels the interface of `_ODERunner` and left a little hack using one of the existing (but unused) specification parameters to swap between the implementations.  I plan to let this sit for a production cycle and then gut the old implementation after the next public release. 

Some benchmarks:
image

This plot shows the impact on the runtime (y-axis) in seconds for the worse scenario (vaccines, but no mandate reimposition) and the no vaccine scenario (no vaccines, but has mandate reimposition) for two metrics: the total runtime in blue and the time spent in ode processing in orange.  As illustrated, the ode portion makes up 40-90% of the runtime depending on the scenario with the current implementations, but is completely trivial with the optimizations (reflecting about a 100x speedup).  

I'll attach some full scale integration test plots here shortly.


src/covid_model_seiir_pipeline/forecasting/specification.py
```diff
@@ -85,6 +85,7 @@ class ScenarioSpecification:
     )
     ALLOWED_SOLVERS = (
         'RK45',
+        'RK45_optimized'
```
miker985, line 88: trailing comma

miker985 (approved):

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
```diff
@@ -458,3 +465,280 @@ def _beta_shift(beta_hat: pd.DataFrame,
...
+    for i in range(n_groups):
+        dy[i * system_size:(i + 1) * system_size] = _seiir_single_group_system(
+            t, y[i * system_size:(i + 1) * system_size], p, n_total, infectious
+        )
+
+    return dy
+
+
+@numba.njit
+def _vaccine_single_group_system(t: float, y: np.ndarray, p: np.ndarray,
```
kathrynline, line 510: Docstring?

kathrynline: (For all of these functions, actually)

src/covid_model_seiir_pipeline/forecasting/model/ode_forecast.py
```diff
@@ -458,3 +465,280 @@ def _beta_shift(beta_hat: pd.DataFrame,
...
+    if s_vaccines:
+        rho_unprotected = v_unprotected_s / s_vaccines
+        rho_protected = v_protected / s_vaccines
+        rho_immune = v_immune / s_vaccines
+    else:
+        rho_unprotected, rho_protected, rho_immune = 0, 0, 0
+    # Actual count of vaccines coming out of S.
+    s_vaccines = min(1 - beta * infectious ** alpha / n_total - theta_plus, s_vaccines / s) * s
+
+    # Expected vaccines coming out of E.
```
kathrynline, line 547: These constants are confusing to me - E, I1, I2, R.

kathrynline (approved):
It would be great to have docstrings eventually, but ok if that is a backburner item.

collijk:
And we have exact numerical matches with the non-optimized outputs now: 
image

collijk:
:point_up: Is the global aggregate.

PR #125, opened 2020-12-11
Rethink data io

collijk:
### Data I/O

I've stuck to the original design, so this is for details.  

---------------------------------------------------
### Description from Draft PR:

I'm preparing to split off post-processing into a separate pipeline stage (and then fold the hospital model into this one), and so this seemed like a good time to unify and cleanup the I/O abstractions.  **I'm looking for design feedback and gotchas** at this point.  If y'all are free sometime tomorrow and want to talk through stuff, do let me know.  

I've sorted the abstractions into two orthogonal layers:

**Format neutral representations of data sets and metadata sets.**

The concrete representation of these things are the ``DatasetKey`` and ``MetadataKey`` classes (which are simple namedtuples) in the `src/covid_model_seiir_pipeline/io/keys.py` module and contain all the information necessary to put or retrieve the data or metadata set from disk.  Names of the fields could still use a think.

These are served to clients interacting with a concrete subclass of a ``DataRoot``, where a ``DataRoot`` subclass instance encodes a concrete date.version directory in one of the main named dirs in /ihme/covid-19 and all of its substructure.  ``DataRoot`` objects are instantiated with a full path to a date.version directory and an on disk format for metadata sets.  

Subdirectory structure is bound to ``DataRoot`` subclasses at class definition time using ``DatasetType`` and ``MetadataType`` instances which serve as factories for the two key types at runtime.  The ``RegressionRoot`` is the simplest version of this to grok.  ``ForecastRoot`` is slightly more complex due to scenario subdirs.  ``CovariateRoot`` has some dynamic binding behavior to support custom covariates in the experimental pipeline without code modification.  ``InfectionRoot`` is a total hack because the directory structure is, well, very poorly structured (but also destined for obsolescence in the next month or so).  

This group of abstractions takes all the substructure encoded in the old ``Paths`` abstractions (throwing away the staticly encoded information about on disk format) and merges it with the interface features of the ``Keys`` abstraction from the marshall code.  

**Stateless I/O strategies**

Because all of the information about how datasets should streamed to disk in terms of format and logical file structure is now encoded in the keys, I've pared away the root and seed management of the original marshalls and turned them into dynamically selected strategies for streaming to disk.  They are unified through a pretty dead simple interface in ``src/covid_model_seiir_pipeline/io/api.py``.

The only major point of deviation is I swapped the h5py implementation with a pandas/pytables one.  This is mainly because I'm more familiar with the latter and required less code modification.  I was able to substitute all the new versions of the marshalls into the existing test suite with very little effort.  Both of you have spent much more time working with hdf datasets in different contexts and actually putting metrics on this stuff, so if this was a mistake, please let me know.

#### Still todo:
- Get better at naming
- Documentation for more of the important stuff.
- A bit more unit testing, especially around the new representational abstractions
- Integration tests.


src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -5,82 +5,62 @@
```
collijk, line 5: Stuff in this file is useful as a preview of a working implementation, but not in need of deep review at this point.

src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -208,32 +179,39 @@ def load_covariates(self, scenario: ScenarioSpecification, location_ids: List[in
...
-        info_df = pd.read_csv(info_path)
+    def load_mobility_info(self, info_type: str, location_ids: List[int]):
+        info_df = io.load(self.covariate_root.mobility_info(info_type=info_type))
+        return self._format_covariate_data(info_df, location_ids)
+
+    def load_vaccine_info(self, info_type: str, location_ids: List[int]):
+        info_df = io.load(self.covariate_root.vaccine_info(info_type=info_type))
+        return self._format_covariate_data(info_df, location_ids)
+
+    def _format_covariate_data(self, dataset: pd.DataFrame, location_ids: List[int], with_observed: bool = False):
```
collijk, line 190: Just cleaning up the above functions with this one.

src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -246,68 +224,51 @@ def load_scenario_specific_data(self,
...
     def save_output_summaries(self, output_summaries: pd.DataFrame, scenario: str, measure: str):
-        self.forecast_marshall.dump(output_summaries,
-                                    key=MKeys.forecast_output_summaries(scenario=scenario, measure=measure))
+        io.dump(output_summaries, self.forecast_root.output_summaries(scenario=scenario, measure=measure))
 
     def save_output_miscellaneous(self, output_miscellaneous: pd.DataFrame, scenario: str, measure: str):
-        self.forecast_marshall.dump(output_miscellaneous,
-                                    key=MKeys.forecast_output_miscellaneous(scenario=scenario, measure=measure))
-
-    def save_reimposition_dates(self, reimposition_dates: pd.DataFrame, scenario: str, reimposition_number: int):
```
collijk, line 303: Dead code

src/covid_model_seiir_pipeline/io/data_roots.py
```diff
@@ -0,0 +1,109 @@
...
+
+class InfectionRoot(DataRoot):
+    metadata = MetadataType('metadata')
+
+    def modeled_locations(self) -> List[int]:
+        """Retrieve all of the location specific infection directories."""
+        return [int(p.name.split('_')[-1]) for p in self._root.iterdir() if p.is_dir()]
+
+    @property
+    def infections(self):
```
collijk, line 45: hack hack hack

src/covid_model_seiir_pipeline/io/data_roots.py
```diff
@@ -0,0 +1,109 @@
...
+    lri_mortality = DatasetType('lri_mortality', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    mask_use = DatasetType('mask_use', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    mobility = DatasetType('mobility', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    pneumonia = DatasetType('pneumonia', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    proportion_over_2_5k = DatasetType('proportion_over_2_5k', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    proportion_under_100m = DatasetType('proportion_under_100m', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    smoking_prevalence = DatasetType('smoking_prevalence', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+    testing = DatasetType('testing', LEAF_TEMPLATES.COV_SCENARIO_TEMPLATE)
+
+    mobility_info = DatasetType('mobility', LEAF_TEMPLATES.COV_INFO_TEMPLATE)
```
collijk, line 71: This goes to naming in the keys module.  I was expecting the first arg to `DatasetType` to 1:1 map to class variables.  Turned out not to be the case due to the layout in the covariate dirs, but also I already had the flexibility to handle it, so 🤷 .

src/covid_model_seiir_pipeline/io/keys.py
```diff
@@ -0,0 +1,124 @@
...
+    DRAW_TEMPLATE: str = 'draw_{draw_id}'
+    MEASURE_TEMPLATE: str = '{measure}'
+    COV_SCENARIO_TEMPLATE: str = '{covariate_scenario}_scenario'
+    COV_INFO_TEMPLATE: str = '{info_type}_info'
+
+
+LEAF_TEMPLATES = __LeafTemplates()
+
+
+class __PathTemplates(NamedTuple):
```
collijk, line 30: This is maybe better interpreted as a prefix template.  Things get parsed as 
``root/path_template/dataset_key.data_type/leaf_template``

collijk: path template always resolves to nested directories (in the current implementation).  dataset_key.datatype will turn into a file for hdf/zip and a terminal subdirectory for csv.  leaf_template is a node in hdf/zip and a file in csv.

src/covid_model_seiir_pipeline/io/keys.py
```diff
@@ -0,0 +1,124 @@
...
+            raise ValueError(f'Invalid leaf_template specification: {leaf_template} for DatasetType {self.name}. '
+                             f'leaf_template must be one of {LEAF_TEMPLATES}.')
+        self.leaf_template = leaf_template
+
+        self.root = root
+        # Disk format validated upstream
+        self.disk_format = disk_format
+
+    def __get__(self, instance: 'DataRoot', owner=None) -> 'DatasetType':
+        return type(self)(self.name, self.leaf_template, self.path_template, instance._root, instance._data_format)
```
collijk, line 92: I am almost certainly misusing the descriptor protocol, but it seems to work pretty nicely.

src/covid_model_seiir_pipeline/io/keys.py
```diff
@@ -0,0 +1,124 @@
...
+        leaf_name = self.leaf_template.format(**key_kwargs)
+        path_name = self.path_template.format(**key_kwargs) if self.path_template else None
+        return DatasetKey(self.root, self.disk_format, self.name, leaf_name, path_name)
+
+    def __repr__(self):
+        return f'{type(self).__name__}({", ".join(["=".join([k, str(v)]) for k, v in self.__dict__.items()])})'
+
+
+class MetadataType:
+    def __init__(self, name: str,
```
collijk, line 107: These only have yaml implementations now and are presumed to be at the top level under the date.version root.

src/covid_model_seiir_pipeline/io/marshall.py
```diff
@@ -0,0 +1,195 @@
...
+        zip_path, node = cls._resolve_key(key)
+        with zipfile.ZipFile(zip_path) as container:
+            with container.open(node) as inf:
+                return pd.read_csv(inf)
+
+    @classmethod
+    def exists(cls, key: DatasetKey) -> bool:
+        zip_path, node = cls._resolve_key(key)
+        with zipfile.ZipFile(zip_path) as container:
+            try:
```
collijk, line 78: Similar logic twice (line 96+) could stand to be abstracted.

tests/test_marshall.py
```diff
@@ -1,46 +1,42 @@
...
+from covid_model_seiir_pipeline.io.marshall import (
     CSVMarshall,
-    Hdf5Marshall,
     ZipMarshall,
-    DtypeMarshall,
+    HDF5Marshall,
 )
 
 
 class MarshallInterfaceTests:
```
collijk, line 12: I like this structure for testing but find it a little tricky to debug when stuff is broken.

src/covid_model_seiir_pipeline/io/data_roots.py
miker985: If I'm reading this correctly you've implemented `infection_root_instance.infections(loc_id, draw_id)` in the weirdest way possible.Why this way?

src/covid_model_seiir_pipeline/io/data_roots.py
```diff
@@ -0,0 +1,109 @@
...
+    vaccine_info = DatasetType('vaccine_coverage', LEAF_TEMPLATES.COV_INFO_TEMPLATE)
```
miker985, line 72: Why is it `mobility_info`/vaccine_info` instead of `mobility`/`vaccine_coverage`?

If your above comment explains this my apologies - i didn't grok that.

src/covid_model_seiir_pipeline/io/marshall.py
```diff
@@ -0,0 +1,195 @@
+from contextlib import contextmanager
```
miker985, line 1: I think you wanted `git mv` here since it appears you moved `marshall.py`, updated a few imports, and then added a few lines at the bottom.

But maybe you changed the middle as well?

src/covid_model_seiir_pipeline/io/keys.py
miker985: What is `owner` for?

src/covid_model_seiir_pipeline/regression/data.py
```diff
@@ -162,24 +147,23 @@ def load_covariates(self, covariates: Iterable[str], location_ids: List[int]) ->
     ############################
 
     def save_beta_param_file(self, df: pd.DataFrame, draw_id: int) -> None:
-        self.regression_marshall.dump(df, key=MKeys.parameter(draw_id))
+        io.dump(df, self.regression_root.parameters(draw_id=draw_id))
 
     def save_date_file(self, df: pd.DataFrame, draw_id: int) -> None:
-        self.regression_marshall.dump(df, key=MKeys.date(draw_id))
+        io.dump(df, self.regression_root.dates(draw_id=draw_id))
```
miker985, line 153: Ok so, if I'm reading this correctly you rejiggered a whole slew of things but it all works about the same right?

IIRC when I wrote the first version of this keeping a reference to e.g., `self.regression_marshall` was necessary because that object needed to  hold on to an open file handle as most of the backends were stateful.

Is it the case that you're still doing that, or does this do a full write/flush to disk each time?

tests/test_marshall.py
```diff
@@ -58,118 +54,52 @@ def assert_no_accidental_overwrites(self, instance, data, key):
...
 class TestCSVMarshall(MarshallInterfaceTests):
     @pytest.fixture
-    def instance(self, tmpdir):
-        return CSVMarshall(pathlib.Path(tmpdir))
+    def regression_root(self, tmpdir):
+        return RegressionRoot(tmpdir, data_format='csv')
+
+    @pytest.fixture
+    def instance(self):
+        return CSVMarshall
```
miker985, line 62: You're returning a class instead of an instance here. Or maybe CSVMarshall is a singleton now?

src/covid_model_seiir_pipeline/io/keys.py
miker985: also welcome to the dark side

src/covid_model_seiir_pipeline/io/data_roots.py
collijk: I was going to see if I could mimic the implementation of the Datatype where `data_root_instance.datatype` returns a callable.  But that's dumb.  Program to the interface is the right move.  I blame lateness.

src/covid_model_seiir_pipeline/io/keys.py
collijk: owner should be `type(instance)` I think.  But only on invocations in a certain context.  I'm cargo culting a bit and just grabbed the default __get__ signature

src/covid_model_seiir_pipeline/regression/data.py
collijk: Yes.  Actual usage is pretty close to the same.  The rejiggering is all about unifying the keys and paths abstraction and shifting the burden of state representation to the key-like objects.  

In this situation we are doing a full flush to disk each time.

tests/test_marshall.py
collijk: CSVMarshall (and the other marshalls) are effectively classes used just as namespaces.  I don't have any mechanisms for hard enforcement of singletons, but the intention is to use the classes directly (all methods are class methods).

src/covid_model_seiir_pipeline/regression/data.py
collijk: My understanding of your original implementation was that it was flushing to disk on `dump` operations as well.  the state it was managing wasn't the file handles themselves, but just the paths.

src/covid_model_seiir_pipeline/io/marshall.py
collijk: Yeah, sorry.  I was not organized about workflow while doing this.  I pulled off a marshall at a time and used the parallel implementations for testing as I was developing.  

The csv and zip marshalls are essentially the same with some minor tweaks to make them stateless.

I swapped the backend for the hdf marshall to tables (see pr description) and so got rid of lots of the manual handling necessary with h5py.

src/covid_model_seiir_pipeline/regression/data.py
miker985: Upon reading the code you're right - it's always flushed each time.

src/covid_model_seiir_pipeline/io/data_roots.py
collijk: made this less dumb

JosephWagner (approved):
Looks like a solid approach to me. I don't see any huge gotchas or issues with switching from h5py to pytables (besides it being a large dependency, but hdf is large itself)

src/covid_model_seiir_pipeline/io/data_roots.py
```diff
@@ -0,0 +1,106 @@
...
+        """Retrieve all of the location specific infection directories."""
+        return [int(p.name.split('_')[-1]) for p in self._root.iterdir() if p.is_dir()]
+
+    def infections(self, location_id: int, draw_id: int):
+        data_type = str([m for m in self._root.glob(f"*_{location_id}")][0])
+        return DatasetKey(
+            root=self._root,
+            disk_format=self._data_format,
+            data_type=data_type,
+            leaf_name=f'draw{draw_id:04}_prepped_deaths_and_cases_all_age.csv',
```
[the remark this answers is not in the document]
collijk: Oh, this actually would break in integration.  Good catch.

src/covid_model_seiir_pipeline/io/keys.py
```diff
@@ -0,0 +1,124 @@
...
+        if leaf_template not in LEAF_TEMPLATES:
+            raise ValueError(f'Invalid leaf_template specification: {leaf_template} for DatasetType {self.name}. '
+                             f'leaf_template must be one of {LEAF_TEMPLATES}.')
+        self.leaf_template = leaf_template
+
+        self.root = root
+        # Disk format validated upstream
+        self.disk_format = disk_format
+
+    def __get__(self, instance: 'DataRoot', owner=None) -> 'DatasetType':
```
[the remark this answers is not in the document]
collijk: A DatasetType

src/covid_model_seiir_pipeline/io/keys.py
collijk: Settled on prefix templates

PR #126, opened 2020-12-15
Collijk/feature/use GitHub actions for ci

collijk:
Delete old docs, swap from travis ci (which doesn't run anymore) to github actions for build and test.


miker985:
image

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
+name: build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
miker985, line 7: is centos not an option?

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+    strategy:
+      matrix:
+        python-version: [3.7] #, 3.8, 3.9]
```
miker985, line 10: is this for speed of execution or does the code not currently run on 3.8/9?

miker985 (approved):

.github/workflows/build.yml
collijk: It was for testing.  I should put these back.

.github/workflows/build.yml
collijk: Unfortunately, no.

PR #129, opened 2020-12-21
Collijk/feature/separate postprocessing stage

collijk:
This PR separates postprocessing into it's own seiir subcommand and adds features to support splicing results with previous versions and aggregations up custom hierarchies.  

Currently, the production workflow looks like:

- run forecast & postprocessing
- generate diagnostics
- evaluate results for locations that need to be spliced
- copy the forecast outputs into a new directory (this is like 250GB and takes a couple of hours)
- manually splice in locations and do aggregations (historically error prone and a good way to delay results publication by several hours, especially as new measures are introduced.  The scripts here are a bunch of copy-paste code.)
- run downstream code for hospital/viz/briefing docs

This PR has the workflow as:

- run forecast
- run postprocessing with no splicing and standard aggregation (modeling hierarchy)
- generate diagnostics
- evaluate results for locations that need to be spliced
- run postprocessing with spliced locations and results hierarchy
- run downstream code for hospital/viz/briefing docs

It reduces process time by about a factor of 8 and disk storage by a factor of 5.  It allows for easy splicing from multiple output versions if necessary, supports custom aggregations if we get client requests, and standardizes the toolkit for aggregation.


src/covid_model_seiir_pipeline/cli.py
```diff
@@ -91,10 +93,8 @@ def regress(run_metadata,
                                          logger, with_debugger)
     app_metadata, _ = main(regression_spec, preprocess_only)
 
-    run_metadata['app_metadata'] = app_metadata.to_dict()
-    run_metadata.dump(run_directory / 'metadata.yaml')
-
-    cli_tools.make_links(app_metadata, run_directory, mark_best, production_tag)
+    cli_tools.finish_application(run_metadata, app_metadata,
```
collijk, line 96: Just pulling in Joe's updates

src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -229,22 +214,18 @@ def save_specification(self, specification: ForecastSpecification):
...
-    def load_raw_covariates(self, scenario: str, draw_id: int):
-        covariates = io.load(self.forecast_root.raw_covariates(scenario=scenario, draw_id=draw_id))
-        covariates['date'] = pd.to_datetime(covariates['date'])
-        return covariates
-
     def load_beta_params(self, draw_id: int) -> Dict[str, float]:
         df = io.load(self.regression_root.parameters(draw_id=draw_id))
         return df.set_index('params')['values'].to_dict()
 
+    def save_ode_params(self, ode_params: pd.DataFrame, scenario: str, draw_id: int):
```
collijk, line 221: Saving beta/theta/etc. separately from the compartments to reduce load time in postprocessing. The task producing final betas was 2x longer than the next longest task.

src/covid_model_seiir_pipeline/forecasting/task/beta_forecast.py
```diff
@@ -189,10 +189,13 @@ def run_beta_forecast(draw_id: int, forecast_version: str, scenario_name: str, *
                                                                 min_wait, last_reimposition_end_date)
 
     logger.info('Writing outputs.')
-    components = components.reset_index()
+    ode_param_cols = components.columns.difference(compartment_info.compartments)
```
collijk, line 192: Saving ode params separately for postprocessing load times.

src/covid_model_seiir_pipeline/forecasting/workflow.py
```diff
@@ -1,6 +1,5 @@
-import itertools
 import shutil
```
collijk, line 1: Forecast got way simpler, which is nice.

src/covid_model_seiir_pipeline/postprocessing/data.py
```diff
@@ -0,0 +1,291 @@
...
+        deaths_version = Path(metadata['death']['metadata']['output_path'])
+        es_inputs = pd.read_csv(deaths_version / 'model_data.csv')
+        es_inputs['date'] = pd.to_datetime(es_inputs['date'])
+        es_inputs = es_inputs.set_index(['location_id', 'date'])
+        cumulative_cases = (es_inputs['Confirmed case rate'] * es_inputs['population']).rename('cumulative_cases')
+        cumulative_deaths = (es_inputs['Death rate'] * es_inputs['population']).rename('cumulative_deaths')
+        cumulative_hospitalizations = (es_inputs['Hospitalization rate'] * es_inputs['population'])
+        cumulative_hospitalizations = cumulative_hospitalizations.rename('cumulative_hospitalizations')
+        # TODO: have ES use and produce a config file for this number.
+        es_shift = -8
```
collijk, line 122: This is new. The full_data csv containing input data doesn't fill missing days, but that makes big trouble for the viz.  We want stuff after ES has filled gaps, but the only data we have has been shifted to prepare for the spline models.  So we unshift them.

src/covid_model_seiir_pipeline/postprocessing/data.py
```diff
@@ -0,0 +1,291 @@
+from pathlib import Path
```
collijk, line 1: Most of this is pulled out of the forecast data inteface.

src/covid_model_seiir_pipeline/postprocessing/model/aggregators.py
```diff
@@ -0,0 +1,124 @@
+"""Postprocessing aggregation strategies."""
```
collijk, line 1: Only new things here are aggregating all the way up a hierarchy instead of just national levels + global.  `git mv` can't always save us.

collijk: There's some code duplication here that could stand to be factored out.  Will do it if I ever have to touch this again.

src/covid_model_seiir_pipeline/postprocessing/model/final_outputs.py
```diff
@@ -0,0 +1,193 @@
+from typing import Any, Callable, List
```
collijk, line 1: No new code in this module.  Just parsed off from postprocessing_lib

src/covid_model_seiir_pipeline/postprocessing/model/loaders.py
```diff
@@ -0,0 +1,125 @@
+import functools
```
collijk, line 1: Moved all the logic in these functions to the data interface.  Now this is a wrapper to support concurrent reads.

src/covid_model_seiir_pipeline/postprocessing/model/resampling.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import pandas as pd
+
+from covid_model_seiir_pipeline.postprocessing.specification import ResamplingSpecification
+
+
+def build_resampling_map(deaths: pd.DataFrame, resampling_params: ResamplingSpecification):
+    cumulative_deaths = deaths.groupby(level='location_id').cumsum().reset_index()
+    cumulative_deaths['date'] = pd.to_datetime(cumulative_deaths['date'])
+    reference_date = pd.to_datetime(resampling_params.reference_date)
```
collijk, line 13: Only new thing here is using a date from configuration instead of the last day of the forecast for draw trimming.

src/covid_model_seiir_pipeline/postprocessing/task/postprocess.py
```diff
@@ -0,0 +1,184 @@
+from argparse import ArgumentParser, Namespace
```
collijk, line 1: I'm surprised `git mv` didn't call this a rename.  Same as old postprocess.py task with added aggregation and splicing features.

src/covid_model_seiir_pipeline/postprocessing/task/resample_map.py
```diff
@@ -0,0 +1,56 @@
+from argparse import ArgumentParser, Namespace
```
collijk, line 1: Nothing new in this module.

src/covid_model_seiir_pipeline/forecasting/data.py
```diff
@@ -229,22 +214,18 @@ def save_specification(self, specification: ForecastSpecification):
...
+
     def save_components(self, forecasts: pd.DataFrame, scenario: str, draw_id: int):
         io.dump(forecasts, self.forecast_root.component_draws(scenario=scenario, draw_id=draw_id))
 
-    def load_components(self, scenario: str, draw_id: int) -> pd.DataFrame:
-        components = io.load(self.forecast_root.component_draws(scenario=scenario, draw_id=draw_id))
-        components['date'] = pd.to_datetime(components['date'])
-        return components.set_index(['location_id', 'date'])
+    def load_components(self, scenario: str, draw_id: int):
+        return io.load(self.forecast_root.component_draws(scenario=scenario, draw_id=draw_id))
```
miker985, line 228: Did you consistently push the date conversion into `io.load`? If not you're asking for trouble IFF a non-typed implementation like CSV ever gets used

collijk: I haven't done that yet.  It is very definitely on my list.

miker985: OK. My admittedly poor memory is that you either have to do a bad heuristic ala `if 'date' in df` or leave it here.

Or just delete the non-typed backends, which is probably reasonable.

collijk: I just realized that this code changed cause I deleted it (it's unused in the actual pipeline), and then added it back in because a test failed.  I'll revert to the original implementation.

collijk: I think csv has to stick around for the final outputs.  I've been thinking about the long term switch.  I'm going to make a little perfomance logger in a PR soon and then do some experiments.  Feather looks like a nice alternative from the benchmarks I've seen.

miker985 (commented):
Your existing PR's are frequently noisy and complex enough that I don't understand the "real" changes beyond whatever textual writeup you might provide.

This is a lot more.

Is there something specific you actually want me to check for?

PR #132, opened 2020-12-23
Collijk/refactor/package layout

collijk:
Things were starting to get crowded at the package level and I want to do a few infrastructure bits next, so this first PR is just a reorg int a `lib` subpackage with tools for building pipeline stages and a `pipeline` subpackage for the pipeline stages. I've tried to be very intentional about import structure:
- All imports in `lib` from other things in `lib` are absolute.
- `lib` subpackages declare a public interface in their __init__ so that they look like modules to clients
- `lib` modules and subpackages are used as namespaces in `pipeline` and `cli` code.
- With the exception of `model` subpackages, all imports in a `pipeline` stage to other items in the same `pipeline` stage are absolute.
- `model` subpackages in pipeline stages declare a public interface in their __init__ so that they look like modules. The intent is that they operate like pipeline stage local libraries.
- Subpackages of `pipeline` declare a public interface in their __init__ for usage by other pipeline stages (typically their specification and their data interface)
- The `pipeline` subpackage itself declares a public interface in its __init__ for usage by the cli (all specifications and all application entry points).

There is one bit of "logic" change here which is to wrap the ihme dependencies so CI doesn't fail at import time.  If the package had more tests, we'd have certainly run into this earlier.


.gitignore
```diff
@@ -14,7 +14,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
```
collijk, line 17: Maybe naming stuff lib is a bad idea?

src/covid_model_seiir_pipeline/lib/utilities.py
```diff
@@ -114,9 +116,9 @@ def get_input_root(cli_argument: Optional[str], specification_value: Optional[st
     return root.resolve()
 
 
-def get_location_metadata(location_specification: Optional[str],
-                          spec_lsvid: Optional[int],
-                          spec_location_file: Optional[str]) -> Tuple[int, str]:
+def get_location_info(location_specification: Optional[str],
```
collijk, line 119: This was a name duplication hidden by a local import.  Whoops.

src/covid_model_seiir_pipeline/lib/ihme_deps.py
```diff
@@ -0,0 +1,35 @@
...
+
+    class C:
+        def __init__(self, *args, **kwargs):
+            raise ModuleNotFoundError(f"No module named '{module_name}'")
+
+
+def _lazy_import_error_function(module_name: str):
+    def f(*args, **kwargs):
+        raise ModuleNotFoundError(f"No module named '{module_name}'")
+
```
miker985, line 18: You're not returning anything in either of your `_lazy` methods

collijk: Very good catch.

src/covid_model_seiir_pipeline/lib/ihme_deps.py
```diff
@@ -0,0 +1,35 @@
...
+    from db_queries import get_location_metadata
+except ModuleNotFoundError:
+    get_location_metadata = _lazy_import_error_function('db_queries')
+
+try:
+    from jobmon.client import Workflow, BashTask
+    from jobmon.client.swarm.executors.base import ExecutorParameters
+    from jobmon.client.swarm.workflow.task_dag import DagExecutionStatus
+    from jobmon.client.swarm.workflow.workflow import WorkflowAlreadyComplete
+except ModuleNotFoundError:
```
miker985, line 30: The way you've written this you have to update the list of imports in the `try` and `except` any time there is change.

Suggest re-writing this so that you have a single list of imports. This could be as easy as updating `_lazy_import_foo` to do this

src/covid_model_seiir_pipeline/lib/workflow/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .specification import (
+    TaskSpecification,
+    WorkflowSpecification,
+)
+from .template import (
+    TaskTemplate,
+    WorkflowTemplate
```
miker985, line 7: Why are you mixing relative an absolute imports between submodules?

src/covid_model_seiir_pipeline/pipeline/postprocessing/main.py
```diff
@@ -21,16 +21,16 @@ def do_postprocessing(app_metadata: cli_tools.Metadata,
...
         if unknown_covariates:
             logger.warning("Some covariates that were modeled have no postprocessing configuration. "
                            "Postprocessing will produce no outputs for these covariates. "
                            f"Unknown covariates: {list(unknown_covariates)}")
 
-        measures = (list(final_outputs.MEASURES)
-                    + list(final_outputs.MISCELLANEOUS)
+        measures = (list(model.MEASURES)
+                    + list(model.MISCELLANEOUS)
                     + list(modeled_covariates.intersection(known_covariates)))
```
miker985, line 34: `itertools.chain`?

miker985:
```python
>>> foo = list('abc')
>>> bar = ['doe', 'ray', 'me']
>>> baz = [1,2,3]
>>> [*foo, *bar, *baz]
['a', 'b', 'c', 'doe', 'ray', 'me', 1, 2, 3]
```

src/covid_model_seiir_pipeline/pipeline/postprocessing/model/final_outputs.py
```diff
@@ -1,14 +1,17 @@
...
+from typing import Any, Callable, List, TYPE_CHECKING
 
 import pandas as pd
 
-from covid_model_seiir_pipeline.postprocessing.data import PostprocessingDataInterface
-from covid_model_seiir_pipeline.postprocessing.model import aggregators, loaders
+from covid_model_seiir_pipeline.pipeline.postprocessing.model import aggregators, loaders
+
+if TYPE_CHECKING:
+    from covid_model_seiir_pipeline.pipeline.postprocessing.data import PostprocessingDataInterface
```
miker985, line 8: Is this to avoid a circular dependency? If not, why are you doing this?

src/covid_model_seiir_pipeline/lib/workflow/__init__.py
collijk: I've only got relative imports in the package inits.  The idea is that they:
- declare the subpackage api (obviously)
- enforce the subpackage structure
- Only know local structure.  Which I was thinking would make them easy to move around as a unit without code churn.  Given that the module use absolute imports, this seems like a dumb idea though.

src/covid_model_seiir_pipeline/pipeline/postprocessing/main.py
collijk: Yes, that's the right solution.

src/covid_model_seiir_pipeline/lib/ihme_deps.py
collijk: Updated and with tests even (though only happy path, really)!

src/covid_model_seiir_pipeline/lib/workflow/__init__.py
collijk: I had a reason, but it's dumb on reflection.

collijk: Fixed

src/covid_model_seiir_pipeline/pipeline/postprocessing/main.py
collijk: Fixed

src/covid_model_seiir_pipeline/pipeline/postprocessing/model/final_outputs.py
collijk: This is contract enforcement.  `model` subpackages are libraries to their respective pipeline stages and may be used anywhere.  The fact that these show up for data interfaces and specifications in lots of the model subpackages points to an incongruity in the API.  Those two object types actually need to be shared across pipeline stages.  After which I can relax this contract.

I added boilerplate comments to all the TYPE_CHECKING blocks to explain them.

miker985 (approved):

PR #136, opened 2020-12-28
Collijk/feature/process monitor

collijk:
This PR introduces a logger with a stripped down interface that can track performance information and a little parser utility to unpack and summarize the performance logs.

Example output from a single run of the beta regression:

```
2020-12-27 18:22:06.030 | covid_model_seiir_pipeline.lib.cli_tools:report:120 - 
Runtime report
===============================
setup               :      0.02
read                :     12.13
transform           :      5.18
compute_ode         :    252.28
compute_regression  :    183.13
write               :      6.07 - {}
```
And then a run of the log parser:
```
$> sparse /ihme/covid-19/seir-regression/latest/logs/jobmon/output/ beta_regression
                         mean    min     max
setup                 0.02584   0.01    0.06
read                 11.59767   6.74   17.87
transform             5.50910   2.30   13.72
compute_ode         247.91060  85.79  489.34
compute_regression  140.89023   9.94  268.76
write                 8.22696   4.21   24.52
```

This will need some extra features to handle scenarios and measures that I'll think about when I get there.


src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/log_parser.py
```diff
@@ -0,0 +1,27 @@
+from collections import defaultdict
```
collijk, line 1: Just a move.  No new code

src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/performance_logger.py
```diff
@@ -0,0 +1,47 @@
+from collections import defaultdict
+import time
+
+from loguru import logger
+
+
+class TaskPerformanceLogger:
```
collijk, line 7: This logger is a little bit dumb.  It does automatic context switching, which means that if log messages go missing, we'll over allocate time to subtasks that aren't correct.  Not super worried about this in practice and doing explicit context switching would double the logging statements.

tests/regression/test_model.py
```diff
@@ -1,5 +1,6 @@
 import numpy
 import pandas
+from pandas.testing import assert_frame_equal
```
collijk, line 3: Exact testing of floats finally broke.  🤷 .  Should have been doing tolerance testing anyway.

src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/performance_logger.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self):
+        self.current_context = None
+        self.current_context_start = None
+        self.times = defaultdict(float)
+
+    def _record_timing(self, context):
+        if context is None:
+            return
```
miker985, line 16: should this ever happen?

src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/performance_logger.py
```diff
@@ -0,0 +1,47 @@
...
+    def warning(self, *args, context=None, **kwargs):
+        self._record_timing(context)
+        logger.debug(*args, **kwargs)
+
+    def report(self):
+        if self.current_context is not None:
+            self.times[self.current_context] += time.time() - self.current_context_start
+        logger.info(
+            "\nRuntime report\n" +
+            "="*31 + "\n" +
```
miker985, line 42: spaces around `*`

src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/performance_logger.py
miker985, line 16: suggestion: move this check to the log API methods

src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/performance_logger.py
```diff
@@ -0,0 +1,47 @@
...
+    def _record_timing(self, context):
+        if context is None:
+            return
+        if self.current_context is None:
+            self.current_context = context
+            self.current_context_start = time.time()
+        else:
+            self.times[self.current_context] += time.time() - self.current_context_start
+            self.current_context = context
+            self.current_context_start = time.time()
```
miker985, line 23: This logic means you need to be careful not to repeat contexts or you'll lose time, right?

collijk: I'll only lose the time of the execution of these three lines as times are cumulative.

PR #139, opened 2021-01-06
Collijk/feature/grid plots

collijk:
This PR adds a new pipeline stage `diagnostics` with a single job type `grid_plots` (more to come shortly).  Right now grid plots take 50-80 minutes to run and work out of an Rstudio session.  This version runs in about 5 minutes using multiprocessing and node-local storage and can run multiple sets of grid plots in parallel.  

There's a lot of boilerplate around setting up a new pipeline stage that can mostly be ignored.


src/covid_model_seiir_pipeline/cli.py
```diff
@@ -212,14 +215,71 @@ def postprocess(run_metadata,
...
+                diagnostics_specification,
+                preprocess_only,
+                output_root, mark_best, production_tag,
+                verbose, with_debugger):
+    cli_tools.configure_logging_to_terminal(verbose)
+
+    diagnostics_spec = DiagnosticsSpecification.from_path(diagnostics_specification)
+
+    outputs_versions = set()
+    for grid_plot_spec in diagnostics_spec.grid_plots:
```
collijk, line 239: This is a little nonstandard since there are multiple versions here to track.

PR #145, opened 2021-01-17
Collijk/refactor/cli tech debt

collijk:
This refactors the cli.py to use some shared functions, which I didn't have time to do when I put in the run_all command.  This significantly reduces code duplication, which is always nice.


kathrynline (approved):
So clean!

PR #152, opened 2021-01-19
Collijk/feature/postprocess hospital

collijk:
Thread hospital data and new extra outputs through postprocessing.

This also cleans up some inconsistencies in the command templates.

### DIFF
--- a/specification_templates/postprocessing.yaml
+++ b/specification_templates/postprocessing.yaml
@@ -1,16 +1,16 @@
 data:
-  forecast_version: '2021_01_13.05'
+  forecast_version: '2021_01_18.01'
   scenarios:
     - 'worse'
     - 'reference'
-    - 'no_vaccine'
-    - 'vaccine_fast'
-    - 'vaccine_fast_elderly'
-    - 'best_masks'
+#    - 'no_vaccine'
+#    - 'vaccine_fast'
+#    - 'vaccine_fast_elderly'
+#    - 'best_masks'
   output_root: ''
 workflow:
   project: 'proj_covid'
-  queue: 'd.q'
+  queue: 'all.q'
   tasks:
     resample_map:
       max_runtime_seconds: 5000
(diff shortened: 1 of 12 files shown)


specification_templates/postprocessing.yaml
```diff
@@ -1,16 +1,16 @@
...
     - 'worse'
     - 'reference'
-    - 'no_vaccine'
-    - 'vaccine_fast'
-    - 'vaccine_fast_elderly'
-    - 'best_masks'
+#    - 'no_vaccine'
+#    - 'vaccine_fast'
+#    - 'vaccine_fast_elderly'
+#    - 'best_masks'
```
miker985, line 9: should these be deleted?

kathrynline (approved):

src/covid_model_seiir_pipeline/pipeline/postprocessing/model/final_outputs.py
```diff
@@ -73,11 +73,63 @@ def __init__(self,
         cumulative_label='cumulative_infections',
         aggregator=aggregators.sum_aggregator,
     ),
+    'r_controlled': MeasureConfig(
+        loaders.load_r_controlled,
+        'r_controlled',
+        splice=False,  # TODO: Get a few rounds of this and then set splice to True
```
miker985, line 79: if all these `splice` args are the same thing then you should set a single variable and reuse it.

miker985 (approved):

src/covid_model_seiir_pipeline/pipeline/postprocessing/model/final_outputs.py
collijk: good call

specification_templates/postprocessing.yaml
collijk: You should ignore churn in the specifications.  I'll set up a gitignore for the ones I use for testing soon.

PR #167, opened 2021-02-07
Make plotting more robust to missing data

collijk:
Trying to reduce how much hacking needs to happen to get plots made in prod.  This is a little gross, but a refactor is for later.


src/covid_model_seiir_pipeline/pipeline/diagnostics/model/plot_version.py
```diff
@@ -74,20 +75,34 @@ def build_cache(self, cache_dir: Path, cache_draws: List[str]):
         (self._cache / 'miscellaneous').mkdir()
 
         for measure in [*MEASURES.values(), *COVARIATES.values()]:
-            summary_data = self.pdi.load_output_summaries(self.scenario, measure.label)
-            summary_data.to_parquet(self._cache / 'summaries' / f'{measure.label}.parquet', compression=None)
+            try:
+                summary_data = self.pdi.load_output_summaries(self.scenario, measure.label)
+                summary_data.to_parquet(self._cache / 'summaries' / f'{measure.label}.parquet', compression=None)
```
miker985, line 80: suggestion try/except/else for situations like these

src/covid_model_seiir_pipeline/pipeline/diagnostics/model/plotter.py
```diff
@@ -344,7 +344,10 @@ def make_time_plot(ax,
                    vlines=(),
                    transform=lambda x: x):
     for plot_version in plot_versions:
-        data = plot_version.load_output_summaries(measure, loc_id)
+        try:
+            data = plot_version.load_output_summaries(measure, loc_id)
+        except FileNotFoundError:  # No data for this version, so skip.
+            continue
```
miker985, line 350: Not worth logging?

miker985 (approved):

src/covid_model_seiir_pipeline/pipeline/diagnostics/model/plot_version.py
collijk: That's a much better idea.

PR #170, opened 2021-02-07
Add composite measures with IFR as first case

collijk:
IFR and other ratios need to be composed from count space measures after splicing and aggregation.  This sets up the abstraction for that and uses the abstractions to produce IFR.


src/covid_model_seiir_pipeline/pipeline/postprocessing/model/final_outputs.py
```diff
@@ -26,6 +26,16 @@ def __init__(self,
...
 
 
+class CompositeMeasureConfig:
+    def __init__(self,
+                 base_measures: Dict[str, MeasureConfig],
+                 label: str,
+                 combiner: Callable = None):
+        self.base_measures = base_measures
+        self.label = label
+        self.combiner = combiner
```
miker985, line 36: Will this object be valid without a combiner?

miker985 (approved):

PR #171, opened 2021-02-07
Collijk/feature/new esi rates regression

collijk:
With the new infection elastispliner, we stopped modeling the hospital fatality ratio directly and started modeling the infection fatality ratio, infection hospitalization ratio, and infection detection ratio instead.  This PR incorporates those changes into the calculation of the hospital correction factors in the regression stage and sets us up to read in the ratios for computing additional metrics in forecasting and postprocessing.


src/covid_model_seiir_pipeline/pipeline/forecasting/model/forecast_metrics.py
```diff
@@ -77,7 +76,7 @@ def compute_output_metrics(infection_data: pd.DataFrame,
 
 def compute_corrected_hospital_usage(all_age_deaths: pd.DataFrame,
                                      death_weights: pd.Series,
-                                     hospital_fatality_ratio: HospitalFatalityRatioData,
+                                     hospital_fatality_ratio,
```
miker985, line 79: You deleted the type, but the argument is still valid?

collijk: It's a series now.  Need to come back and update typing.

PR #178, opened 2021-02-15
Collijk/feature/push index handling down the stack

collijk:
This is just a first pass as pushing index column handling down the stack.  Still need to construct internal contracts about what's in the index vs not at what layers and make sure everything is consistent about it.  That's to do later.


kathrynline (approved):

specification_templates/postprocessing.yaml
```diff
@@ -26,9 +26,9 @@ resampling:
   reference_date: '2021-12-31'
   lower_quantile: 0.025
   upper_quantile: 0.975
-splicing:
-  - locations: [11, 123, 133, 527, 532, 555, 85, 198, 197, 195, 196, 193, 194, 184]
-    output_version: '/ihme/covid-19/seir-outputs/2021_02_03.02'
+#splicing:
+#  - locations: [11, 123, 133, 527, 532, 555, 85, 198, 197, 195, 196, 193, 194, 184]
+#    output_version: '/ihme/covid-19/seir-outputs/2021_02_03.02'
```
miker985, line 31: these commented lines should be cleaned up

specification_templates/regression.yaml
```diff
@@ -18,7 +18,7 @@ workflow:
       m_mem_free: '20G'
       num_cores: 26
 regression_parameters:
-  n_draws: 1000
+  n_draws: 100
```
miker985, line 21: seems dangerous - is this the new number of draws?

src/covid_model_seiir_pipeline/lib/io/marshall.py
```diff
@@ -10,6 +10,8 @@
     MetadataKey,
 )
 
+POTENTIAL_INDEX_COLUMNS = ['location_id', 'date', 'age_start']
```
miker985, line 13: Normally I would tell you constants should be tuples, not lists

But pandas is sometimes sensitive to types and tuples won't work, so instead I ask - can this be a tuple?

src/covid_model_seiir_pipeline/pipeline/diagnostics/task.py
```diff
@@ -73,6 +73,7 @@ def run_grid_plots(diagnostics_version: str, name: str, progress_bar: bool) -> N
             date_end=pd.to_datetime(grid_plot_spec.date_end),
             output_dir=plot_cache,
         )
+        _runner(locs_to_plot[0])
```
miker985, line 76: wat?

src/covid_model_seiir_pipeline/pipeline/forecasting/data.py
```diff
@@ -71,21 +71,16 @@ def load_location_ids(self) -> List[int]:
     def load_regression_coefficients(self, draw_id: int) -> pd.DataFrame:
         return self._get_regression_data_interface().load_regression_coefficients(draw_id=draw_id)
 
-    def load_transition_date(self, draw_id: int) -> pd.Series:
+    def load_transition_date(self, draw_id: int) -> pd.MultiIndex:
         dates_df = self._get_regression_data_interface().load_date_file(draw_id=draw_id)
         dates_df['end_date'] = pd.to_datetime(dates_df['end_date'])
-        transition_date = dates_df.set_index('location_id').sort_index()['end_date'].rename('date')
-        return transition_date
+        return dates_df['end_date'].rename('date')
```
miker985, line 77: Is a DataFrame a MultiIndex? I thought the latter was a type of index but fundamentally more like a Series than a DataFrame

or is dates_df not really a DataFrame?

src/covid_model_seiir_pipeline/pipeline/forecasting/data.py
```diff
@@ -71,21 +71,16 @@ def load_location_ids(self) -> List[int]:
...
 
     def load_beta_regression(self, draw_id: int) -> pd.DataFrame:
-        beta_regression = self._get_regression_data_interface().load_regression_betas(draw_id=draw_id)
-        beta_regression['date'] = pd.to_datetime(beta_regression['date'])
-        return beta_regression
+        return self._get_regression_data_interface().load_regression_betas(draw_id=draw_id)
```
miker985, line 80: I don't think you can sanely make these changes unless you guarantee the data you're loading is loaded from a typed file format like parquet.

If CSV is still available you should be using a method in `pandas.api.types` to check if you need to convert via `pd.to_datetime`

src/covid_model_seiir_pipeline/pipeline/forecasting/model/forecast_metrics.py
```diff
@@ -49,7 +49,7 @@ def compute_output_metrics(infection_data: pd.DataFrame,
     past_infections_idx = components_past.set_index('date', append=True).index
     modeled_infections = modeled_infections.to_frame()
     modeled_deaths = modeled_deaths.reset_index(level='observed')
-    infection_data = infection_data.set_index(['location_id', 'date'])
+    infection_data = infection_data
```
miker985, line 52: image

src/covid_model_seiir_pipeline/pipeline/forecasting/task/beta_forecast.py
```diff
@@ -42,10 +42,10 @@ def run_beta_forecast(forecast_version: str, scenario: str, draw_id: int):
     # Covariates and coefficients, and scaling parameters are
     # used to compute beta hat in the future.
     covariates = data_interface.load_covariates(scenario_spec)
-    coefficients = data_interface.load_regression_coefficients(draw_id)
+    coefficients = data_interface.load_regression_coefficients(draw_id).reset_index()
     forecast_end_date = covariates.date.max()
     # Rescaling parameters for the beta forecast.
-    beta_scales = data_interface.load_beta_scales(scenario=scenario, draw_id=draw_id)
+    beta_scales = data_interface.load_beta_scales(scenario=scenario, draw_id=draw_id).reset_index()
```
miker985, line 48: These two changes suggest to me that `load_foo` methods should all reset the index before returning their result. thoughts?

src/covid_model_seiir_pipeline/lib/io/marshall.py
collijk: It can be! I think I was using it directly at first, but I'm just iterating over it in current usage.

src/covid_model_seiir_pipeline/pipeline/diagnostics/task.py
collijk: Validation leftovers.  Hard to debug multiprocessing code.

src/covid_model_seiir_pipeline/pipeline/forecasting/data.py
collijk: Swapped to index and updated typing, then I realized swapping to an index was a bad idea and reverted, but did not fix the typing.

src/covid_model_seiir_pipeline/pipeline/forecasting/data.py
collijk: The CSV marshall is looking for a 'date' column and converting before we get to this layer.  Any other date-like columns have to be handled explicitly (as is done in `load_transition_date` above).  Is there something else you're worried about?  Or do you think that's unsafe?

src/covid_model_seiir_pipeline/pipeline/forecasting/model/forecast_metrics.py
collijk: 🤦

src/covid_model_seiir_pipeline/pipeline/forecasting/task/beta_forecast.py
collijk: I'm leaving it up at this level.  I introduce an index abstraction in a later PR in the forecast stage.  I want the noise here at this level to annoy me into backporting it into regression soon.

src/covid_model_seiir_pipeline/pipeline/forecasting/data.py
miker985: If this one was always redundant than removing it is dandy. I noticed a **lot** of other similar removals, I didn't pay any attention to the column being renamed.

I perceive the risk of you loading CSV data to be extremely low. Nonetheless you've introduced change that could lead to an unintended regression were that to happen. This clashes with my "use types and do other things very carefully" hat which is why I brought it up.

PR #181, opened 2021-02-15
Collijk/feature/parquet format regression outputs

collijk:
Log parser is from #180.  Can be ignored for the purposes of this PR.

This only impacts the `write` step of regression.  It improves both write times and file size on disk.

| Version | Total size (GB) | Total write time (s) | Average write time (s) | 
| :---            |                ---: | ---: | ---: |
| csv           |             46.0  | 7653.9 | 7.65 |
| parquet    |             12.1  | 2709.3 | 2.71 |

### DIFF
--- a/src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/log_parser.py
+++ b/src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/log_parser.py
@@ -24,4 +24,7 @@ def parse_logs(log_dir, job_name):
             time, *_ = time.split('\x1b')
             result[metric.strip()].append(float(time.strip()))
     df = pd.DataFrame(result)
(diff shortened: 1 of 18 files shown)


src/covid_model_seiir_pipeline/lib/cli_tools/performance_logger/log_parser.py
```diff
@@ -24,4 +24,7 @@ def parse_logs(log_dir, job_name):
             time, *_ = time.split('\x1b')
             result[metric.strip()].append(float(time.strip()))
     df = pd.DataFrame(result)
-    click.echo(df.describe().T[['mean', 'min', 'max']])
```
kathrynline, line 27: Didn't you make this same change in the previous PR?

kathrynline (approved):

miker985 (approved):

PR #183, opened 2021-02-15
Collijk/feature/parquet format forecast outputs

collijk:
Both read from regression and write output in parquet format

| Version             | Total size (GB) | Total time (s) | Average time (s) | 
| :---                     |                    ---: |                ---: |                       ---: |
| read (csv)          |                N/A   |         711269 |                   118.5 |
| read (parquet)   |                N/A   |      1017534  |                  169.6  |
| write(csv)          |             2197.1 |         634084 |                   105.7 | 
| write(parquet)   |               606.7 |     303282     |                   50.55 |

### DIFF
--- a/specification_templates/forecast.yaml
+++ b/specification_templates/forecast.yaml
@@ -1,12 +1,12 @@
 data:
-  regression_version: '2021_02_15.05'
-  covariate_version: '2021_02_11.01'
+  regression_version: '2021_02_18.05'
+  covariate_version: '2021_02_18.01'
   output_root: ''
   output_format: 'parquet'
   fh_subnationals: False
 workflow:
   project: 'proj_covid_prod'
-  queue: 'all.q'
+  queue: 'd.q'
   tasks:
     beta_residual_scaling:
       max_runtime_seconds: 5000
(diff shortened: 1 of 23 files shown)


specification_templates/forecast.yaml
```diff
@@ -1,15 +1,16 @@
...
+  output_format: 'parquet'
   fh_subnationals: False
 workflow:
   project: 'proj_covid_prod'
   queue: 'all.q'
   tasks:
     beta_residual_scaling:
       max_runtime_seconds: 5000
-      m_mem_free: '5G'
+      m_mem_free: '50G'
```
miker985, line 13: I'm all for throwing memory at the problem but I have to ask - did this actually cause a 10x memory usage increase or did you just pick a big number so it wouldn't oom?

If so, what's the real number?

src/covid_model_seiir_pipeline/pipeline/forecasting/specification.py
```diff
@@ -45,6 +45,7 @@ class ForecastData:
     regression_version: str = field(default='best')
     covariate_version: str = field(default='best')
     output_root: str = field(default='')
+    output_format: str = field(default='csv')
```
miker985, line 48: What are you thoughts on emitting a warning or some other log message if this default is used?

miker985 (approved):

specification_templates/forecast.yaml
collijk: On my to do list.  Parquet does compression and so need more mem to load from disk.  Just 10xed it for now since it's unlikely for me to find 26 free threads without 50G of mem on a node.

src/covid_model_seiir_pipeline/pipeline/forecasting/specification.py
collijk: I think a deprecation warning is a good idea.

PR #185, opened 2021-02-15
Index abstraction and construction

collijk:
This introduces an abstraction around indices to make it easier to keep datasets square.  All the data prep code and results processing code will use this in the new system.  Eventually needs to be backported to regression, but the new system is the urgent thing to do right now.

### DIFF
--- a/src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
+++ b/src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
@@ -17,6 +17,63 @@ from covid_model_seiir_pipeline.pipeline.regression.model.containers import (
 )
 
 
+class Indices:
+    """Abstraction for building square datasets."""
+
+    def __init__(self,
+                 regression_start_dates: pd.Series,
+                 forecast_start_dates: pd.Series,
+                 forecast_end_dates: pd.Series):
+        self._past_index = self._build_index(regression_start_dates, forecast_start_dates, pd.Timedelta(days=1))
+        self._future_index = self._build_index(forecast_start_dates, forecast_end_dates)
+        self._initial_condition_index = (
+            forecast_start_dates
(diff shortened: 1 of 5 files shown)


src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
```diff
@@ -17,6 +17,63 @@
 )
 
 
+class Indices:
+    """Abstraction for building square datasets."""
```
miker985, line 21: I'm familiar with the colloquial use of "square" but I'm confused as to what that means in Pandas given that all `DataFrame`s are square by definition (or maybe I'm mistaken and they aren't? they always provide a square `.shape`...)

src/covid_model_seiir_pipeline/pipeline/forecasting/task/beta_forecast.py
```diff
@@ -24,6 +24,25 @@ def run_beta_forecast(forecast_version: str, scenario: str, draw_id: int):
...
+    # unique datasets in this model and they need to be aligned consistently
+    # to do computation.
+    logger.info('Loading index building data', context='read')
+    regression_start_dates = data_interface.load_regression_start_dates(draw_id)
+    forecast_start_dates = data_interface.load_forecast_start_dates(draw_id)
+    # Forecast is run to the end of the covariates
+    covariates = data_interface.load_covariates(scenario_spec)
+    forecast_end_dates = covariates.reset_index().groupby('location_id').date.max()
+    logger.info('Building indices', context='transform')
+    indices = model.Indices(
```
miker985, line 40: so it's intentional that these `indices` are not yet used by the "old system".

Why not delay creating the instance until you're ready to use it? Just to see if creating it errors?

miker985 (approved):

PR #187, opened 2021-02-15
Isolate seiir model parameter prep

collijk:
This PR builds an abstraction around the model parameters, ensuring they have consistent indices.

### DIFF
--- a/src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
+++ b/src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
@@ -75,6 +75,54 @@ class Indices:
 
 
 @dataclass
+class ModelParameters:
+    # Core parameters
+    alpha: pd.Series
+    beta: pd.Series
+    sigma: pd.Series
+    gamma1: pd.Series
+    gamma2: pd.Series
+
+    # Theta parameters
+    theta_plus: pd.Series
+    theta_minus: pd.Series
+
+    # Vaccine parameters
+    unprotected_lr: pd.Series
+    protected_wild_type_lr: pd.Series
+    protected_all_types_lr: pd.Series
+    immune_wild_type_lr: pd.Series
+    immune_all_types_lr: pd.Series
+    old_unprotected_lr: pd.Series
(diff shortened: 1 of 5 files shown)


src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
```diff
@@ -74,6 +74,54 @@ def _build_index(start: pd.Series,
...
+    old_immune_hr: pd.Series
+
+    # Variant parameters
+    beta_b117: pd.Series
+    beta_b1351: pd.Series
+    b117_prevalence: pd.Series
+    b1351_prevalence: pd.Series
+    probability_cross_immune: pd.Series
+
+    def with_index(self, index: pd.MultiIndex):
```
miker985, line 116: This is more like `at_index` or `subset_to_index` than it is `with_index`

Perhaps to pedantic, but it seems like now would be the time...

src/covid_model_seiir_pipeline/pipeline/forecasting/model/ode_forecast.py
```diff
@@ -33,16 +35,176 @@
...
+        probability_cross_immune=probability_cross_immune,
+    )
+
+
+def adjust_vaccinations_for_variants(vaccine_data: pd.DataFrame, covariates: pd.DataFrame):
+    vaccine_data = vaccine_data.reindex(covariates.index, fill_value=0)
+    if 'variant_prevalence_B1351' in covariates.columns:
+        b1351_prevalence = covariates['variant_prevalence_B1351']
+    else:
+        b1351_prevalence = pd.Series(0, index=vaccine_data.index)
```
miker985, line 95: Is this something you'd want to log/warn for? If this branch ever happened by accident it'd probably save you diagnostic/debug time...

miker985 (approved):
Mostly I don't understand what the changes are doing, but I assume you already ran it so it's not going to error

If you haven't, I'd suggest running `flake8` or somesuch to check for unused variables. Enough churn happened in some of the larger functions it seems conceivable that it would pick up something.

PR #190, opened 2021-02-16
Collijk/feature/new ode runner

collijk:
Draft of a new layout for handling multiple ode systems.  I don't love how split parameters/compartments/etc are over this and the ode systems, but that's a problem for a different day.


src/covid_model_seiir_pipeline/pipeline/forecasting/model/ode_forecast.py
```diff
@@ -345,125 +344,121 @@ def forecast_correction_factors(indices: Indices,
...
+                model_parameters.protected_wild_type_lr,
+                model_parameters.protected_all_types_lr,
+                model_parameters.immune_wild_type_lr,
+                model_parameters.immune_all_types_lr,
+                model_parameters.unprotected_hr,
+                model_parameters.protected_wild_type_hr,
+                model_parameters.protected_all_types_hr,
+                model_parameters.immune_wild_type_hr,
+                model_parameters.immune_all_types_hr,
+            ]
```
miker985, line 430: At this point you should consider using `type()` with 3 arguments.

miker985 (approved):

PR #191, opened 2021-02-16
Collijk/feature/new output metrics

collijk:


src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
```diff
@@ -158,22 +161,60 @@ def to_dict(self) -> Dict[str, Union[int, pd.Series, pd.DataFrame]]:
...
     herd_immunity: pd.Series
-    total_susceptible: pd.Series
-    total_immune: pd.Series
 
-    def to_dict(self) -> Dict[str, pd.DataFrame]:
+    def to_dict(self) -> Dict[str, pd.Series]:
         return utilities.asdict(self)
 
+    def to_df(self) -> pd.DataFrame:
+        return pd.concat([v.rename(k) for k, v in self.to_dict().items()], axis=1)
```
miker985, line 216: Random question: why not `return pd.DataFrame(self.to_dict())`?

miker985 (approved):

src/covid_model_seiir_pipeline/pipeline/forecasting/model/containers.py
collijk: Because I'm not always clever.

PR #192, opened 2021-02-16
Collijk/feature/column selection from marshalled data

collijk:


src/covid_model_seiir_pipeline/lib/io/marshall.py
```diff
@@ -46,26 +46,26 @@ def touch(cls, *paths: Path) -> None:
...
-    def _resolve_key(cls, key: DatasetKey) -> Path:
+    def _resolve_key(cls, key: DatasetKey) -> Tuple[Path, Union[List[str], None]]:
         path = key.root
         if key.prefix:
             path /= key.prefix
         path /= key.data_type
         if key.leaf_name:
             path /= key.leaf_name
-        return path.with_suffix(".csv")
+        return path.with_suffix(".csv"), key.columns
```
miker985, line 60: How do you know the columns actually exist?

tests/lib/io/test_keys.py
```diff
@@ -95,7 +95,7 @@ class MyDataRoot(DataRoot):
     expected = MetadataType('metadata', mdr._root, mdr._metadata_format)
     assert mdr.metadata == expected
 
-    expected_key = DatasetKey(mdr._root, mdr._data_format, 'saturn', 'draw_0', 'cassini')
+    expected_key = DatasetKey(mdr._root, mdr._data_format, 'saturn', 'draw_0', 'cassini', None)
```
miker985, line 98: I assume `None` is the default which returns all columns.

IMO you should omit this explicit `None` and have `DatasetKey` supply a default value.

```python
>>> class Keys(typing.NamedTuple):
...   name: str
...   columns: typing.Optional[str] = None                                                                                                                                                                                                     
...                                                                                                                                                                                                                                            
>>> k1 = Keys(name="data")
>>> k2 = Keys(name="other data", columns=['foo', 'bar'])                                                                                                                                                                                       
>>> k1
Keys(name='data', columns=None)                                                                                                                                                                                                                
>>> k2                                                                                                                                                                                                                                         
Keys(name='other data', columns=['foo', 'bar'])
```

src/covid_model_seiir_pipeline/lib/io/marshall.py
collijk: I don't, but pandas raises an uncharacteristically reasonable error if you ask for columns that don't exist.

